PR #2, opened 2021-06-17
feat(multiselect): Added multiselect - FRONT-2351

tibi2303:


src/components/select/select.story.js
```diff
@@ -13,6 +14,17 @@ const applyArgs = (data, args) => {
   return Object.assign(data, args);
 };
 
+const withInit = (story) => {
```
papegaill, line 17: The following controls doesn't work:
- Invalid, can we add invalid style like the simple select?
- Required, maybe remove the control in this case?
- Input size

papegaill: Maybe we should plan a twig `placeholder` param to be able to change the default "select value" for the multi-select

src/components/select/select.story.js
```diff
@@ -13,6 +14,17 @@ const applyArgs = (data, args) => {
...
+  const demo = story();
+  return `
+    ${demo}
+    <script>        
+      new SlimSelect({
+        select: '.multi-select'
+      })
```
papegaill, line 24: it may be good to add the option `selectByGroup: true`, and have a optgroup example in the options?

storybook/.storybook/preview-head.html
```diff
@@ -3,6 +3,12 @@
   type="text/javascript"
   src="oe-bcl-default.bundle.min.js"
 ></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/slim-select/1.27.0/slimselect.min.js"></script>
+<link
+  rel="stylesheet"
+  type="text/css"
+  href="https://cdnjs.cloudflare.com/ajax/libs/slim-select/1.27.0/slimselect.min.css"
+/>
```
papegaill, line 11: I think we should more investigate on the styles, sass? in the current state, we have no possibility to change the styles. We need to match as much as possible default bootstrap style and colors. can we somehow map primary colors or badge styles here?

tibi2303: I found a solution using js on DOMNodeInserted when an option is selected. We need to discuss how we deliver this external component js.

src/components/select/select.story.js
tibi2303: I extended and added this functionality at select component level.

src/components/select/select.story.js
tibi2303: Actually invalid and required works, it's adding the according attributes, Indeed i removed input size and placeholder.

src/components/select/select.story.js
```diff
@@ -13,6 +14,17 @@ const applyArgs = (data, args) => {
   return Object.assign(data, args);
 };
 
+const withInit = (story) => {
+  const demo = story();
+  return `
+    ${demo}
```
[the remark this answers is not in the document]
tibi2303: I couldn't find a way to add this, there is no doc regarding this in package https://storybook.js.org/addons/@whitespace/storybook-addon-html . I was looking for something like disable js on that tab.

src/components/select/select.story.js
emeryro: The attributes are added in the code indeed, but there is no visual indication of it (like red or grey color). Are there some options in the third part library to make it more visible?

tibi2303: I added the error state in multiselect.scss.

tibi2303: Seems like slim select does not provide this states, I made the custom error state to make it look like the bootstrap one and I think we can overwrite the default styles in multiselect.scss

storybook/.storybook/preview-head.html
```diff
@@ -3,6 +3,12 @@
   type="text/javascript"
   src="oe-bcl-default.bundle.min.js"
 ></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/slim-select/1.27.0/slimselect.min.js"></script>
```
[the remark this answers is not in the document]
planctus: In general, i would avoid bundling external libs in the js we provide, but at the same time if it's usage is not optional we probably should.
In this case the script would then be available, also in storybook, through the bundle.min.js file of the theme and it would be matter of importing it in the js index files.
If we instead decide that this is optional we could still use the npm package, but we would need to copy that in the asset folder for storybook to use it.

src/components/select/select.story.js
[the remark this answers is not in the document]
planctus: this is related to the comment above, the thing is that if this is a functionality which comes by default with the library this should be a module and added to the js we bundle.
If it's not, and this is just for the sake of the demo then this might be probably moved in the preview-head, although i see that you refer to something coming from the data (but i guess that maybe is not needed, it could be hardocoded), in this case the script tag would not be "part" of the story and the html tab would not show it.

storybook/.storybook/preview-head.html
planctus: mmh..i don't necessarily understand the approach here, i think it would probably be better, if we want to override some of the styles of this library to use their sass files which defines a few variables that can be overridden:

```
$height: 30px !default;
$white: #ffffff !default;
$font-color: #666666 !default;
$font-placeholder-color: #8a8a8a !default;
$font-disabled-color: #dedede !default;
$primary-color: #5897fb !default;
$border-color: #dcdee2 !default;
$search-highlight-color: #fffb8c !default;
$border-radius: 4px !default;
$spacing-l: 8px !default;
$spacing-m: 6px !default;
$spacing-s: 4px !default;
```

we can have or scss file "using" the scss from slimselect and override what we need more than adding custom js and custom css for making slight changes, i believe..

src/components/select/select.story.js
planctus: mmh...in the comment below i added some considerations that might also affect this..

storybook/.storybook/preview-head.html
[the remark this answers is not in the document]
planctus: Seeing the implementation here, i think that in the end any kind of usage of this "component" might be different, based on specific needs, i honestly feel this could stay as an "optional" library to be added, for which we should provide documentation on how to use it, but i don't necessarily see the opportunity to bundle this, which would make the js heavier while it might be only needed in one single page and that could be handled at the application level

planctus:
There were some misunderstandings about the previous review, i guess:

- i think that we should use the npm package of slimselect also because we should provide the js (not in the bundle) in the theme packages.
- the npm package also contains the scss file, the idea is to overwrite just what is needed, preferrably using the variables provided by the slimselect.scss
- it's ok to move the call in the js in the preview-head.html but we need to check if we are in the right story...

In the end i've tested myself these changes, i've pushed a branch on top of yours:` chore-using-npm-package-slimselect`, where the things i mentioned are implemented, i'm not enthusiast about this approach either and i leave to you the choice about what to take from there, you can merge it if you want, or cherry-pick only what you "like".
Mind that in that branch i haven't reproduced the css overrides you had, i just wanted to show you what i intended when i said we could "use" the `slimselect.scss`, in case you go for that, you have to report there the customisations you made, but again, especially if it's minor things i would limit as much as possible the "overrides".
Then, this has been done only for the default theme, i really don't know if it's worth applying the same to the "ecl" one, in any case it looks to me that we might have to create a "base" theme with all the common things at some point and then the "themes" would add their own specific things.

papegaill:
As agreed, I merge this PR and the styling aspects will be taken into consideration by another ticket, thank you @tibi2303

papegaill (approved):

PR #4, opened 2021-06-25
chore(attributes): Using drupal attributes in the alert component

planctus:
I've found an existing implementation of the drupal attributes in twing made by the same author, it seems to work as close as possible to the drupal implementation, as it is, in general, for how twing implements twig-php.
I made an example on the alert component, we would then refactor all the components to apply this new approach.


tibi2303:
Pretty nice, I should start on this refactoring. Should i include it in this PR?

tibi2303 (approved):

PR #5, opened 2021-06-29
feat(button-drupal): Refactor button using drupal attributes - FRONT-2371

tibi2303:


src/components/button/button.html.twig
```diff
@@ -37,36 +37,50 @@
...
+{% set attributes = attributes.removeAttribute('class') %}
+{% set attributes = attributes.removeAttribute('aria-current') %}
+{% set attributes = attributes.removeAttribute('aria-expanded') %}
+{% set attributes = attributes.removeAttribute('autocomplete') %}
+{% set attributes = attributes.removeAttribute('data-bs-toggle') %}
+{% set attributes = attributes.removeAttribute('id') %}
+{% set attributes = attributes.removeAttribute('disabled') %}
+{% set attributes = attributes.removeAttribute('aria-disabled') %}
+{% set attributes = attributes.removeAttribute('aria-controls') %}
+{% set attributes = attributes.removeAttribute('data-bs-target') %}
```
tibi2303, line 50: This is done because of the include of buttons inside loops. It persists the attributes from the previous items so I needed to reset them somehow. I don't like this approach but I couldn't find a way to reset attributes to a default null value (it's a map object in js but in twig i can't reset it).

src/components/button/button.html.twig
```diff
@@ -37,36 +37,50 @@
...
-{% set _class = _class ~ (_text_nowrap ? ' text-nowrap' : '') %}
 {% if _extra_classes is not empty %}
-  {% set _class = (_class ? _class ~ ' ' : '') ~ _extra_classes %}
+  {% set _class = _class|merge([_extra_classes]) %}
+{% endif %}
+{% if _id is not empty %}
+  {% set attributes = attributes.setAttribute('id', _id) %}
+{% endif %}
+{% if _disabled %}
+  {% set attributes = attributes.setAttribute('disabled', '').setAttribute('aria-disabled', 'true') %}
```
tibi2303, line 65: Seems like for single attribute with only a name we still need to provide it a value, even if it's empty. I searched a bit and i don't think we can add single prop attributes like disabled or readonly without a value prop.

src/components/button/button.story.js
```diff
@@ -146,6 +147,7 @@ const applyArgs = (data, args) => {
   if (args.name == "none") {
     data.icon = null;
   }
+  data.attributes = new drupalAttribute();
```
tibi2303, line 150: we need still to provide this object type in each twig/story where we include a button, if we don't, the attributes don't work because the package installed does not get in action because we don't actually access the story which access the button twig, we only access the twig for the include.

tibi2303: If we can find a way to determine that the attributes variable is actually a drupalAttribute only in twig, that would fix everything.

src/components/navigation/navigation.html.twig
```diff
@@ -103,9 +103,13 @@
       >
       {% endif %}
       {% if _item.dropdown is defined %}
-        {% include '@oe-bcl/dropdown/dropdown.html.twig' with _item only %}
+        {% include '@oe-bcl/dropdown/dropdown.html.twig' with _item|merge({
+          attributes: attributes
```
tibi2303, line 107: we need to add this attributes also for twig files that include the button twig, it's a domino effect.

src/components/button/button.html.twig
planctus: if that's the case, better to set the booleans as `true`, not an empty value, probably.
But...the thing is that this `attributes` is supposed to entirely replace the `extra_attributes` and `extra_classes` parameters we currently have, the idea is that when implemented in `drupal` all the needed attributes will be passed the "drupal" way, so as `attributes`.

Also the idea of having a parameter for setting an attribute in the template seems to me questionable, the `id` for instance will be provided as an item in `attributes` directly, no..?
There are surely cases where we might want to add attributes via the template, the `disabled` parameter for instance is used for setting two attributes, at least, but still it looks to me like this is not the job of the template anymore, we have to provide examples but we will probably simplify the life of the developers by simply removing parameters when they are only going to be used to set attributes.

src/components/button/button.html.twig
planctus: But this is because of how we are attaching the attributes from the `drupalAttribute()` class to the items, right?
The problem is not twig and resetting the attributes in twig but to feed each include of the template with a new implementation of `drupalAttribute()`, i believe, or the one with the right attributes.
Now, the thing is that this might all start from the specs, we could and probably should attach attributes directly from the "source", this would simulate the "real" world scenario, also because many times we would pre-populate this object.
This way we would not have to process this in the stories and we would not have to do it again in the tests, to recap:

1. extra_attributes and extra_classes should simply disappear
2. parameters used to set attributes should probably disappear, not saying it's a rule but in most cases i believe they should
3. The best place for attaching attributes to our items is probably from the specs, we do it once and we are done with them.
4. since we will remove many parameters used to set attributes, we would set those attrs in the specs using `drupalAttribute`
5. Not sure of how much you want to keep from this first implementation you did, i might suggest restarting fresh more than undoing many of the changes you made.

EX:
in the button group specs you could import drupalAttribute():
`const drupalAttribute = require("drupal-attribute");`
and then attach the attributes to each item:
```
      {
        input: true,
        label: "I am a input field",
        toggle_variant: "outline-primary",
        type: "checkbox",
        toggle: true,
        attributes: new drupalAttribute().setAttribute('id', 'submit-1'),
      },
```

src/components/navigation/navigation.html.twig
planctus: No, this is actually the reason for some of the issues you experienced, i think, why would you use the "navigation" attributes for the templates it includes..? Now i understand that this "task" of refactoring the button is complex, there are so many templates including it, maybe if you want we could actually start from a simpler example, where you can control better what is happening and getting some confidence about it, still i think it might have been useful to better understand what the idea is, we are not simply "using" a function mimicking drupal, we are replacing most of the logic the template had to set attributes for the root of the template.

PR #8, opened 2021-07-05
feat(drupal-attributes): Drupal attributes for multiple components - FRONT-2408

tibi2303:
Components refactored modal,icon,blockquote,spinners,table with drupal attributes

### DIFF
--- a/src/components/blockquote/blockquote.html.twig
+++ b/src/components/blockquote/blockquote.html.twig
@@ -13,9 +13,6 @@
 {% set _cite = cite|default('') %}
 {% set _alignment = alignment|default('left') %}
 {% set _classes = ['text-' ~ _alignment ] %}
-{% if attributes is empty %}
-  {% set attributes = create_attribute() %}
-{% endif %}
 
 {% set attributes = attributes.addClass(_classes) %}
 
--- a/src/components/blockquote/blockquote.html.twig
+++ b/src/components/blockquote/blockquote.html.twig
@@ -12,8 +12,16 @@
 {% set _attribution = attribution|default('') %}
 {% set _cite = cite|default('') %}
 {% set _alignment = alignment|default('left') %}
+{% set _classes = ['text-' ~ _alignment ] %}
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
 
-<figure class="text-{{ _alignment }}">
+{% set attributes = attributes.addClass(_classes) %}
+
+<figure
+  {{ attributes }}
+>
   <blockquote class="blockquote">
     <p>{{- _quote -}}</p>
   </blockquote>
--- a/src/components/blockquote/blockquote.test.js
+++ b/src/components/blockquote/blockquote.test.js
@@ -2,26 +2,26 @@ import { merge, renderTwigFileAsNode } from "@openeuropa/bcl-test-utils";
 import demoData from "@openeuropa/bcl-data-blockquote/data";
 
 const template = "@oe-bcl/blockquote/blockquote.html.twig";
-const render = (params) => renderTwigFileAsNode(template, params);
+const render = (params, reset) => renderTwigFileAsNode(template, params, reset);
 
 describe("OE - Blockquote", () => {
   test(`renders correctly`, () => {
     expect.assertions(1);
 
-    return expect(render({ ...demoData })).resolves.toMatchSnapshot();
+    return expect(render({ ...demoData }, true)).resolves.toMatchSnapshot();
   });
 
   test(`renders correctly with no cite`, () => {
     expect.assertions(1);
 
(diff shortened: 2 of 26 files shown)


src/components/blockquote/blockquote.html.twig
```diff
@@ -12,8 +12,16 @@
 {% set _attribution = attribution|default('') %}
 {% set _cite = cite|default('') %}
 {% set _alignment = alignment|default('left') %}
+{% set _classes = ['text-' ~ _alignment ] %}
+{% if attributes is empty %}
```
planctus, line 16: mmhh..why this? Attributes is always available for the template in drupal and there is no reason why we would need to create a new one. The case when we need to add it programmatically is when we need to pass a "drupal attributes" map into an included template, otherwise for us is a matter of making it available to the template from the specs.

src/components/icon/icon.html.twig
```diff
@@ -7,25 +7,26 @@
...
-{% set _class = 'bi ' ~ 'icon--' ~ _size %}
+{% set _classes = ['bi', 'icon--' ~ _size] %}
 
 {% if _transformation is not empty %}
-  {% set _class = _class ~ ' icon--' ~ _transformation %}
+  {% set _classes = _classes|merge(['icon--' ~ _transformation]) %}
 {% endif %}
-{% if _extra_classes is not empty %}
-  {% set _class = _class ~ ' ' ~ _extra_classes %}
+{% if attributes is empty %}
```
planctus, line 21: Same as before, this is the main "attributes", available for all the twig templates in drupal.

src/components/spinner/spinner.html.twig
```diff
@@ -6,26 +6,31 @@
...
+  {% set _classes = _classes|merge(['text-' ~ _variant]) %}
 {% endif %}
-{% if _extra_classes is not empty %}
-  {% set _class = _class ~ ' ' ~ _extra_classes %}
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
 {% endif %}
 
-<div class="{{ _class }}" role="status">
+{% set attributes = attributes
```
planctus, line 26: i think you can put this in one line, i don't really like the style we are using but when we have multiple attributes i agree to put theme in a different line, here it seems it can be just a single line, no..?

planctus (changes requested):

src/components/blockquote/blockquote.test.js
```diff
@@ -2,34 +2,36 @@ import { merge, renderTwigFileAsNode } from "@openeuropa/bcl-test-utils";
 import demoData from "@openeuropa/bcl-data-blockquote/data";
 
 const template = "@oe-bcl/blockquote/blockquote.html.twig";
-const render = (params) => renderTwigFileAsNode(template, params);
+const render = (params, reset) => renderTwigFileAsNode(template, params, reset);
```
planctus, line 5: you only need to pass it to renderTwigFileAsNode once, @tibi2303,
`const render = (params) => renderTwigFileAsNode(template, params, true);`
what you did might only be needed in case for some reasons, one of the tests would have a different need, but i hardly see that happening..

src/components/button/button.html.twig
```diff
@@ -56,6 +56,12 @@
     {% endif %}
   {% endfor %}
 {% endif %}
+{% if _icon is not empty and _icon.attributes is empty %}
+  {% set _icon = _icon|merge({
+      attributes: create_attribute()
```
planctus, line 61: but why..? The attributes will be available in drupal in the icon template, if we need to add them for our own "demos" this has to happen in the specs, The use case where we need to use create_attribute() is when we are setting attributes in the include code, basically...

src/components/link/link.html.twig
```diff
@@ -26,9 +26,15 @@
 {% set _class = '' %}
 
 {% if _icon is not empty %}
-{% set _icon_markup %}
-  {% include '@oe-bcl/icon/icon.html.twig' with _icon only %}
-{% endset %}
+  {% if _icon.attributes is empty %}
```
planctus, line 29: Same as before, i don't understand why in twig we should do (for drupal) something that drupal does by default...

src/components/modal/modal.html.twig
```diff
@@ -6,16 +6,24 @@
...
   - footer (block)
-  - extra_class
 #}
 {% set _size = size ? 'modal-' ~ size : '' %}
 {% set _id = id|default('defaultModalId') %}
 {% set _header = header|default('') %}
 {% set _body = body|default('') %}
 {% set _footer = footer|default('') %}
-{% set _extra_class = extra_class|default('') %}
+{% set _class = ['modal fade'] %}
```
planctus, line 15: can this be `_classes` everywhere..?

planctus (approved):

PR #9, opened 2021-07-06
feat(drupal-link): Refactor link component - FRONT-2417

tibi2303:


src/components/breadcrumb/breadcrumb.html.twig
```diff
@@ -14,6 +14,11 @@
   <ol class="{{ _class }}">
     {% for _link in _links %}
       {% if _link.path %}
+        {% if _link.attributes is empty %}
```
planctus, line 17: But why...? We only need to do this if we are setting in the template some attributes, otherwise we are creating something that on drupal is available by default..

src/components/dropdown/dropdown.html.twig
```diff
@@ -55,74 +55,57 @@
...
-{% set _trigger_extra_classes = 'dropdown-toggle' %}
-{% if _trigger.extra_classes is not empty %}
-  {% set _trigger_extra_classes = _trigger_extra_classes ~ ' ' ~ _trigger.extra_classes %}
+{% if _trigger.attributes is empty %}
+  {% set _trigger = _trigger|merge({
+    attributes: create_attribute()
+  }) %}
 {% endif %}
+{% set _trigger = _trigger|merge({
+  id: _id,
```
planctus, line 64: couldn't we pass this as an setAttribute as well instead of doing it via the custom parameter..?

src/components/link/link.story.js
```diff
@@ -64,6 +64,7 @@ const applyArgs = (data, args) => {
   if (args.name == "none") {
```
planctus, line 64: in the story shouldn't we link the collapse with the collapse docs in bootstrap..?
https://getbootstrap.com/docs/5.0/components/collapse/

planctus (changes requested):
It seems pretty good, i've mainly looked at the code, but should be fine, just a few comments..

planctus (approved):

PR #10, opened 2021-07-07
feat(drupal-attributes): Refactored multiple components - FRONT-2419

tibi2303:
Refactored components using drupal attributes: accordion, badge, list-group, dropdown, offcanvas.

### DIFF
--- a/src/components/badge/badge.story.js
+++ b/src/components/badge/badge.story.js
@@ -38,7 +38,7 @@ const getArgTypes = (data) => {
     background: {
       type: { name: "select" },
       description: "Background color of the badge",
-      defaultValue: data.background,
+      defaultValue: "primary" || data.background,
       control: {
         type: "select",
         options: getVariants(),
--- a/src/components/badge/badge.story.js
+++ b/src/components/badge/badge.story.js
@@ -73,6 +73,9 @@ const getArgTypes = (data) => {
 };
 
 const applyArgs = (data, args) => {
+  data.attributes.removeClass(`bg-${data.background}`);
+  data.attributes.removeClass("rounded-pill");
+
   return Object.assign(data, args);
 };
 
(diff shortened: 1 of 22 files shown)


src/components/dropdown/dropdown.story.js
```diff
@@ -26,7 +26,7 @@ const getArgTypes = (data) => {
     dark: {
       type: { name: "boolean" },
       description: "Enables dark mode for dropdown menu",
-      defaultValue: data.toolbar,
+      defaultValue: "false",
```
planctus, line 29: boolean `false`, here is a string...

src/components/badge/badge.story.js
```diff
@@ -73,6 +73,9 @@ const getArgTypes = (data) => {
 };
```
planctus, line 73: can we fix this small thing in the badge story in this pr..? The background control should have a value` 'primary' || data.background` since the default is set in the template...

src/components/list-group/list-group.story.js
```diff
@@ -42,6 +42,9 @@ const getArgTypes = () => {
 };
```
planctus, line 42: can we also fix this issue in the list-group demo..? when we switch to horizontal the badge has zero spacing..

planctus (dismissed):
It seems all good to me, we have the chance to fix these small issues we have in some demos but the refactoring is fine.. ;)

src/components/list-group/list-group.story.js
tibi2303: done.

src/components/badge/badge.story.js
tibi2303: done.

src/components/list-group/list-group.story.js
tibi2303: Done

planctus (approved):

PR #11, opened 2021-07-08
feat(drupal-attributes): Refactored select, input, form with drupal attributes - FRONT-2428

tibi2303:


src/components/form-input/form-input.story.js
```diff
@@ -42,7 +42,21 @@ const getArgTypes = (data, type) => {
   };
```
planctus, line 42: can we fix the toggle_variant control..? Should have the default value` 'primary' || data.variant`.. ;)

planctus (approved):

src/components/form-input/form-input.story.js
tibi2303: done

src/components/form/form.html.twig
```diff
@@ -23,12 +23,9 @@
```
planctus, line 23: can you remove the extra_classes also from the doc comment in the template?

tibi2303: done

planctus (approved):

PR #12, opened 2021-07-08
feat(drupal-attributes): Refactored breadcrumb, card, card-layout, button-group - FRONT-2430

tibi2303:


src/components/breadcrumb/breadcrumb.html.twig
```diff
@@ -2,16 +2,18 @@
...
 
-{% set _extra_classes = extra_classes|default('') %}
-{% set _class = 'breadcrumb' ~ (_extra_classes ? ' ' ~ _extra_classes) %}
+{% set _classes = ['breadcrumb'] %}
 {% set _links = links|default([]) %}
 
+{% set attributes = attributes.addClass(_classes) %}
+
 {% if _links is not empty %}
 <nav aria-label="breadcrumb">
```
planctus, line 13: mmh...not that is mandatory that attributes is set on the root element of the component, but here i would expect it to be acting on the nav tag, not the ol...I see that we were using extra_classes setting them on the ol, but also that i think it was "wrong"

src/components/card-layout/card-layout.html.twig
```diff
@@ -48,9 +48,8 @@
 {% if _items is not empty and _items is iterable %}
   {% for _item in _items %}
     {% if _equal_height %}
-      {% set _item_extra_classes = _item.extra_classes ~ ' h-100' %}
       {% set _item = _item|merge({
-          extra_classes: _item_extra_classes
+          attributes: _item.attributes.addClass('h-100')
```
planctus, line 52: it seems here you need to create the attributes (if they are not set) for each item, twig would not have any "attributes" to add a class to, otherwise. You probably set this in the specs, but it should be twig in this case to create them..

src/components/card/__snapshots__/card.test.js.snap
```diff
@@ -101,7 +101,7 @@ exports[`OE - Card renders horizontal correctly 1`] = `
 exports[`OE - Card renders with border variant correctly 1`] = `
 <jest>
   <div
-    class="card border-danger w-25"
```
planctus, line 104: Is this correct..? just asking

src/data/card-layout/dataGroup.js
```diff
@@ -22,6 +24,7 @@ module.exports = {
       card_footer: "Footer of card 1",
       wrapper_class: "col",
       content: "<button class='btn btn-primary'>Custom action</button>",
+      attributes: new drupalAttribute(),
```
planctus, line 27: indeed, this would be correct in case we were not using the attributes in the "parent" template, it would have been just making the attributes available in the included template like it's going to be in Drupal, but since we use the object, this is not needed here and it has to be done in the template, instead.

src/components/card/__snapshots__/card.test.js.snap
tibi2303: as i'm resetting the drupal attributes for each test, it resets the addClass from data, it will happen only in tests so I think is correct this way.

src/components/breadcrumb/breadcrumb.html.twig
tibi2303: The extra_classes was removed and the attributes were added on root element.

src/components/card-layout/card-layout.html.twig
tibi2303: I added the check of attributes outside the equal_height if.

src/data/card-layout/dataGroup.js
tibi2303: It was removed.

planctus (approved):

PR #23, opened 2021-07-15
feat(drupal-attributes): Refactored reset of drupal attributes - FRONT-2472

tibi2303:


src/components/bcl-badge/badge.story.js
```diff
@@ -73,8 +74,7 @@ const getArgTypes = (data) => {
 };
 
 const applyArgs = (data, args) => {
-  data.attributes.removeClass(`bg-${data.background}`);
-  data.attributes.removeClass("rounded-pill");
+  data.attributes = new drupalAttribute();
 
   return Object.assign(data, args);
 };
```
planctus, line 80: can we fix the assistive_text control...? is currently named `hidden_text` and obviously it doesn't work..

src/components/bcl-button-group/button-group.story.js
```diff
@@ -82,10 +83,13 @@ const getArgTypes = (data) => {
...
+};
+
 const applyArgs = (data, args) => {
-  data.attributes
-    .removeClass("btn-group-" + data.size)
-    .removeClass("btn-group-vertical");
+  resetAttrs(data);
+
   return Object.assign(data, args);
 };
```
planctus, line 94: can we fix the `role` control defaultValue (we will have to remove all these defaultValue, it is deprecated...), its value is `group` and should probably be `data.role || 'group'`

src/components/bcl-button/button.story.js
```diff
@@ -123,20 +123,13 @@ const getArgTypes = (data) => {
...
-  data.attributes.removeClass(`btn-outline-${data.variant}`);
-  if (!args.disabled) {
-    data.attributes.removeAttribute("disabled");
-    data.attributes.removeAttribute("aria-disabled");
-  }
-  if (!args.text_nowrap) {
-    data.attributes.removeClass("text-nowrap");
-  }
+const resetAttrs = (data) => {
+  data.attributes = new drupalAttribute();
```
planctus, line 127: no, it looks like we are messing up...clearly this approach doesn't work when we have attributes set at the data level, this way we reset them.
I think we could probably go back to the previous approach, just that we have to check whether the `data.attributes `exists and only if it doesn't we add it.

planctus (changes requested):
We need to go back one step, i believe, the idea of just resetting them by creating a new empty attribute was a bit simplistic. 
Still the approach we had before trying this easy solution was ok, imho, just that we need to add the attributes if they are not in place before resetting what we need.

planctus (approved):

PR #26, opened 2021-07-22
feat(image structure): Added parse image in carousel for drupal structure of images - FRONT-2591

tibi2303:
This was done to keep the alignment with the drupal image structure.

### DIFF
--- a/src/components/bcl-carousel/carousel.story.js
+++ b/src/components/bcl-carousel/carousel.story.js
@@ -127,6 +127,19 @@ const getArgTypes = (data) => {
   };
 };
 
+const parseImages = (data) => {
+  data.items.forEach((slide) => {
+    if (slide.image) {
+      let alt = ``;
+      if (slide.image.alt) {
+        alt = `alt="${slide.image.alt}"`;
+      }
+      slide.image = `<img src="${slide.image.src}" ${alt} class="d-block w-100">`;
+    }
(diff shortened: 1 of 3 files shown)


papegaill (approved):

src/components/bcl-carousel/carousel.story.js
```diff
@@ -127,6 +127,19 @@ const getArgTypes = (data) => {
...
 };
 
+const parseImages = (data) => {
+  data.items.forEach((slide) => {
+    if (slide.image) {
+      let alt = ``;
+      if (slide.image.alt) {
+        alt = `alt="${slide.image.alt}"`;
+      }
+      slide.image = `<img src="${slide.image.src}" ${alt} class="d-block w-100">`;
```
papegaill, line 137: Not sure that these classes will be passed with the Drupal images object `d-block` `w-100`, How you will handle this on theme side?

tibi2303: using the attributes in the yml, but i think they are not necessary in the theme, i think only in the demo storybook. Right now they are passed from yml file

papegaill: ok let's keep it like that, I would like to avoid having to pass attributes on the drupal side for each images .. but we'll see if we need to improve this in the future.

PR #28, opened 2021-07-22
feat(button): Add path param to get a button as a link - FRONT-2599

papegaill:


src/components/bcl-link/link.html.twig
```diff
@@ -5,6 +5,10 @@
   - path (string) (default: '#')
   - variant (string) (default: '')
     - options ["primary", "secondary", "success", "danger", "warning", "info", "light", "dark"]
+  - btn_styled (boolean) (default: false)
```
planctus, line 8: Mmhhh...i don't really think we should add these parameters to the link component, basically we use attributes for this reason, not to multiply the parameters in the template, here is a matter of passing certain attributes to it to get the desired result, which can be done without altering anything.

src/components/bcl-link/link.html.twig
```diff
@@ -35,9 +42,13 @@
   {% set _classes = _classes|merge(['disabled']) %}
   {% set attributes = attributes.setAttribute('aria-disabled', 'true').setAttribute('tabindex', '-1') %}
 {% endif %}
-{% if _variant is not empty %}
+{% if _btn_styled == false and _variant is not empty %}
   {% set _classes = _classes|merge(['link-' ~ _variant]) %}
 {% endif %}
+{% if _btn_styled %}
```
planctus, line 48: Beside the fact that this is unneeded for me, we could receive this already as attributes when you want to use a link and make it look like a button, here we do something strange, we add the class `btn`, then a class `btn-` which only in case there is an outline becomes` btn-outline- ` and only in case the variant is set it becomes an usable css class as `btn-outline-{variant}`. Then, separately, we add another class.
We don't have to use escaping while setting attributes values, they are already going to be escaped by the attributes method used.
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Template%21Attribute.php/class/Attribute/8.5.x

src/components/bcl-link/link.story.js
```diff
@@ -48,6 +48,40 @@ const getArgTypes = (data) => {
         category: "Style",
       },
     },
+    btn_styled: {
```
planctus, line 51: As said, to me these parameters are unneeded, but if you really want to keep them tests should be added for them, they would show the current inconsistencies.

planctus (changes requested):

src/components/bcl-link/link.html.twig
papegaill: if you pass `variant` you will have the `link-variant` class, so to me it can be confusing to k ow how to generate a button link, only by passing attributes..

src/components/bcl-link/link.html.twig
papegaill: I just took this part of the button component code..

papegaill: I just took this part of the button component code..

src/components/bcl-link/link.html.twig
papegaill: So your idea would be to leave the component as is and probably put an explanation that to generate the style of a button you can pass the `btn`, `btn-..`, `btn-outline- ..` classes as attributes?

planctus: To me the opposite is confusing, both when we were thinking of altering the button template to output a link and now that we are adding button params to a link.
I would actually say that the idea of using a link that it looks like a button is confusing per se:
https://medium.com/simple-human/but-sometimes-links-look-like-buttons-and-buttons-look-like-links-9b371c57b3d2

Still the possibility is there, you can do it with the current template and it is just a matter of passing classes.
The idea that i have is that, to a certain extent, we should not have parameters anymore if they are only meant to add attributes, since we use the drupal attributes, but the agreement with the "drupal guys" was to leave things as they were, in the first phase, and then see if we could simplify the templates by removing some or all those params, so i just feel i don't want to increase their number, cause i actually feel we could go decreasing them in the future :)

PR #31, opened 2021-07-27
feat(modal): Add missing bootstrap params - FRONT-2686

papegaill:


src/components/bcl-modal/modal.html.twig
```diff
@@ -26,10 +29,21 @@
                     .addClass(_classes)
 %}
 
+{% if _static_backdrop %}
```
planctus, line 32: I don't think it's particularly convenient here to use drupal attributes here, basically we do that when we include a template and therefore we can expect some attributes coming from drupal, here we can simply "statically" define them.

But anyway, for how you are setting the attributes here, the static_backdrop parameter will never set the relative attribute, the dialog_attributes (it should be a local variable in case you want to keep using it) gets reset right after, indeed.

I think we should also add test for the new parameters, and in this case that would show that this data-bs-backdrop would not be set.

src/components/bcl-modal/modal.html.twig
```diff
@@ -7,10 +7,13 @@
   - body (block)
   - footer (block)
 #}
-{% set _size = size ? 'modal-' ~ size : '' %}
-{% set _id = id|default('defaultModalId') %}
+{% set _size = size ? 'modal-' ~ size %}
```
planctus, line 10: I find this a bit inconsistent now, shouldn't we do the same as you did for the new parameter, our local variable should only be a "copy" of the "public" one and then checking for its value we would set the associated class.

src/components/bcl-modal/modal.html.twig
```diff
@@ -26,10 +29,21 @@
                     .addClass(_classes)
 %}
 
+{% if _static_backdrop %}
+  {% set dialog_attributes = dialog_attributes.setAttribute('data-bs-backdrop', 'static') %}
+{% endif %}
+
+{% set dialog_attributes = create_attribute().addClass([
```
planctus, line 36: With this approach i see that we are adding an empty white space in one test, we normally create the array of classes so that it doesn't contain empty ones and then pass it to addClass(). Still mind the previous comment, might not be needed to use drupal attributes here, but it doesn't harm if we do it correctly.

src/components/bcl-modal/modal.html.twig
```diff
@@ -7,10 +7,13 @@
   - body (block)
   - footer (block)
 #}
-{% set _size = size ? 'modal-' ~ size : '' %}
-{% set _id = id|default('defaultModalId') %}
+{% set _size = size ? 'modal-' ~ size %}
+{% set _id = id|default('modal' ~ random(1000)) %}
```
planctus, line 11: mmh..not so important, but can we add a dash before the random numbering..? `modal-`

src/components/bcl-modal/modal.html.twig
```diff
@@ -7,10 +7,13 @@
...
 {% set _header = header|default('') %}
 {% set _body = body|default('') %}
+{% set _static_backdrop = static_backdrop|default(false) %}
```
planctus, line 14: It looks like these new parameters  have not been added in the doc block at the top..

src/components/bcl-modal/modal.story.js
```diff
@@ -26,6 +26,36 @@ const getArgTypes = () => {
         },
       },
     },
+    static_backdrop: {
```
planctus, line 29: We normally use "human readable" labels for the controls, not necessarily the parameter name when it contains underscores, so here you would add a `name` property.

src/components/bcl-modal/modal.story.js
```diff
@@ -26,6 +26,36 @@ const getArgTypes = () => {
...
+      type: { name: "boolean" },
+      description: "Modal will not close when clicking outside it",
+      defaultValue: false,
```
planctus, line 32: In another pull request these defaultValue has been removed everywhere and rewritten in a different way following the deprecation in storybook.
I would merge that since it affects all the components, once that is done you would use the same approach for the definition of the defaultValue(s) of the controls, ok?

planctus (changes requested):

src/components/bcl-modal/modal.html.twig
papegaill: Thanks for the static_backdrop, it's fixed now. but I don't see why we just have to limit the use of attributes for elements that should receive attributes. personally I find it cleaner than making a condition for each class. I also corrected the extra empty space.

src/components/bcl-modal/modal.html.twig
papegaill: fixed

src/components/bcl-modal/modal.story.js
papegaill: I don't see the point here, can you clarify? There is plenty of underscore parameter in the stories.. cards, button, etc..

src/components/bcl-modal/modal.story.js
papegaill: Waiting to merge that PR ..

src/components/bcl-modal/modal.html.twig
papegaill: Ok, I changed the approach as you suggested, thanks for recommandation!

src/components/bcl-modal/modal.story.js
papegaill: fixed

src/components/bcl-header/__snapshots__/header.test.js.snap
```diff
@@ -238,7 +238,7 @@ exports[`OE - Header ECL renders correctly 1`] = `
     tabindex="-1"
   >
     <div
-      class="modal-dialog modal-fullscreen"
```
planctus, line 241: I think you forgot the size in the template :)
But, for consistency i would use the same approach as in the other attrs, define the local variable as "empty" in case it is not set and then when you "define" the class when you create it.

planctus (changes requested):

src/components/bcl-modal/modal.test.js
```diff
@@ -11,4 +11,36 @@ describe("OE - modal", () => {
...
+
+  test("renders correctly with verticaly centered modal", () => {
+    expect.assertions(1);
+
+    return expect(
+      render({ ...demoData, verticaly_centered: true })
+    ).resolves.toMatchSnapshot();
+  });
+
+  test("renders correctly with verticaly centered modal", () => {
```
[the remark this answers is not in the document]
papegaill: yes 👍

planctus (approved):

PR #32, opened 2021-07-28
chore: Multiple projects POC

planctus:


docs/developing.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Architecture of the library
+
+The BCL library is the tool meant to develop all the components and compositions
+to be used in a drupal application for creating the desired look and feel for a
+specific project.
+It is based on the concept of `themes` which can be directly used in a drupal
+theme providing it with all the needed resources the library has: js files,
+css files, twig templates.
+All the components and compositiions are using bootstrap markup and the default
```
papegaill, line 11: compositi**i**ons

docs/developing.md
```diff
@@ -0,0 +1,159 @@
...
+#### Bootstrap
+
+The library is heavily based on bootstrap, the default components are basically
+twig implementations of the components as defined by bootstrap.
+There shouldn't be any need of using the bootstrap sources in an application
+when using a theme provided by the library but in any case a package with all the
+bootstrap sources is released and maintained by the BCL library.
+The package name is `@openeuropa/bcl-bootstrap` and the folder where it is
+defined in the library is `bootstrap`.
+The library uses Bootstrap 5, the demoes are all having links to the bootstrap
```
papegaill, line 52: demoes should be demos ?

docs/developing.md
```diff
@@ -0,0 +1,159 @@
...
+This usage of the library is discouraged but still possible.
+
+#### Components
+
+Components are stored in `src/components`, in a folder with the name of the
+package that it contains 
+Ex: `src/components/bcl-accordion/accordion.html.twig`.
+These are basically the components for how they are defined by bootstrap, each
+folder contains one or more `*.html.twig` templates, a `*.story.js` file defining
+the demoes that storybook will show, a `*.test.js` file defining test scenarios
```
papegaill, line 68: same here demos?

docs/developing.md
```diff
@@ -0,0 +1,159 @@
...
+might only contain an html markup to be used as the model to be reproduced by
+the application using it.
+Being part of the library, though, they can be written in `twig` if that is
+convenient for the developer, in any case it is not foreseen the usage of those
+twig templates in the drupal application and it is not needed their release as
+npm packages.
+
+#### Stories
+
+Stories are the demoes that storybook will present in the styleguide, the
```
papegaill, line 93: same?

papegaill (commented):
NICE 👍 👍 Working fine and the approach seems good to me! Just a few very small remarks in the readme. The readme is long, and I understand that there is enough to explain, but it may be possible to be shorter in general? what do you think?

docs/developing.md
```diff
@@ -0,0 +1,159 @@
...
+`stories` folder inside the `storybook` folder of each theme, so that the
+developer can freely combine default and custom stories in the project's
+styleguide.
+
+### Overriding and adding twig templates in a theme
+
+All the templates available in `src/components/` are copied in the `templates`
+folder of each theme so that the npm package released will also contain them.
+But these can be overridden, as well as others can be added by placing them in
+the `custom-templates` folder available in each theme.
```
papegaill, line 121: Maybe we can have this chapter also in the empty readme folder of each project?

docs/developing.md
```diff
@@ -0,0 +1,159 @@
...
+##### Combining stories in the theme packages
+
+All the default stories are used by default by all the different themes, but
+they can also be cherry-picked in order to combine them with different stories
+defined by the single projects.
+There is one file in each theme, in the `src/themes/{theme}/storybook/stories`
+folder, called `bcl-stories.js` where a list of the stories to be excluded can
+be provided.
+This folder is also the place where to store the custom stories defined by the
+project, storybook will read then stories from the `bcl-stories` folder and the
```
emeryro, line 111:
```suggestion
project, storybook will read the stories from the `bcl-stories` folder and the
```

docs/release.md
```diff
@@ -0,0 +1,75 @@
...
+```sh
+yarn update-version patch
+```
+or for a new minor:
+
+```sh
+yarn update-version minor
+```
+
+If you need to jump to a higher version number (e.g. if you're skipping some ECL versions, upgrading from ECL 2.3.x to 2.6.y), you will have to provide the version number manually:
```
emeryro, line 21:
```suggestion
If you need to jump to a higher version number (e.g. if you're skipping some BCL versions, upgrading from BCL 2.3.x to 2.6.y), you will have to provide the version number manually:
```

emeryro (commented):
Looks good to me, good job!

A few remarks/questions:
1. It could be good to have an example of template override and/or custom component, similar to what has been done for the scss
2. I assume that if we override a component's twig in a theme, all twig file using this sub component have to be overridden too, to include the proper file, right?

docs/developing.md
```diff
@@ -0,0 +1,159 @@
...
+All the components and compositions are using bootstrap markup and the default
+styles and behaviours, unless amended or overridden by the theme, are provided
+by the bootstrap library, its sources are compiled by the library in a way as
+close as possible to the one used by bootstrap itself.
+
+#### The monorepo
+
+BCL is a `monorepo`, both in terms of the fact that it holds multiple npm
+packages as well as multiple "websites" in a single repository.
+All the packages have then the same versioning and each release happens at the
```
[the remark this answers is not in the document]
planctus: sure, we knew this since the beginning, but if it's true that it's a constraint it's also a way to standardize something that otherwise would have surely tricky to handle.

There is another issue that we might face in the long term, right now is a single "build" and "check" about the preview, once the number of projects will increase significantly this will make the builds time consuming and an error in one single project would block the entire build. We can look in the future for ways to limit the build (in the custom branches) only to the projects affected by the changes..

About the template override the logic is simply that the theme package will ship the one defined by it in its templates folder, the overriding template is supposed to have the same name as the one being overridden, so nothing will change in the other templates about its inclusion. But i'm thinking that at the moment this would not work in our environment...i've tried the template override but only in terms of the distribution of the file, not the usage in our storybook...i'll need to make some changes for this to be possible.. :)

tibi2303 (approved):

PR #34, opened 2021-08-04
feat(param): Renamed type to input_type in form_input - FRONT-2687

tibi2303:


planctus (changes requested):
I think we miss something here, i don't see anything changed in the story where we have the control, indeed the selection of different input types is not working in the preview.
Also the search composition data file should be updated, i guess..

planctus:
It's not so important because we always explicitely set the value for the control. but anyway, what it used to be the `defaultValue `of the control is now the key of the args object that we define in `getArgs()`, `type` should be renamed in `input_type` also there...

planctus (approved):

PR #40, opened 2021-08-09
feat(multiselect): Restyling of multiselect - FRONT-2592

tibi2303:
first we need to add the removable badge and see if we can adapt the functionality of removing a option using our icon.


.storybook/preview-head.html
```diff
@@ -14,6 +14,16 @@
           selectByGroup: true,
           placeholder: "Please select a value",
         });
+        var ssValues = document.querySelectorAll('.ss-values')[0];
```
planctus, line 17: I feel we are running in circle, this cannot be the way, we cannot ask our users to define this custom javascript on their own for each of theirs multiple selects.
If you want to do this you have to bundle a js module that, based on something, like an attribute, retrieve all the multiselects and for each of them does this work.

But my question is..? Would it be worth doing something like this? Following comments might look like an answer.

src/themes/default/src/scss/_multiselect.scss
```diff
@@ -12,3 +12,60 @@
...
+      background-repeat: no-repeat;
+      border-color: #dc3545;
+    }
+  }
+  .badge {
+    margin: 0.25rem 0.45rem 0 0;
+    &:last-child {
+      margin-bottom: 0.25rem;
+    }
+    .ss-value-delete {
```
planctus, line 34: we are back to the situation we had with the former implementations of the "dismissible" badge.
But the thing is, is it really important than instead of the x that slimselect is already setting we override it to have "another" x in this elements? Is it an important, relevant improvement of the design of the multiselect, i don't think so and in any case this was not working for the badges, it's not working here either.

src/themes/default/src/scss/_multiselect.scss
```diff
@@ -12,3 +12,60 @@
...
+        width: 12px;
+        height: 12px;
+        display: inline-block;
+        color: #0d6efd;
+        font-weight: 500;
+      }
+    }
+  }
+  .ss-multi-selected {
+    min-height: 1.95rem;
```
planctus, line 53: There is something i don't understand, the initial request was to try to customise, if needed, the slimselect.scss file by passing different values for the variables defined in the by it in our multiselect scss, one of them is "height" and we don't alter it but we override it here to set a min-height that is less than 2px taller than the default one...

src/themes/default/src/scss/_multiselect.scss
```diff
@@ -12,3 +12,60 @@
...
+    .ss-add {
+      margin-right: 1rem;
+    }
+    .ss-plus {
```
planctus, line 57: if we really had to customise something i believe this should have been jiding this element when the select is invalid.

src/themes/default/src/scss/_multiselect.scss
```diff
@@ -12,3 +12,60 @@
...
+      height: 1rem;
+      top: -3px;
+      &:after {
+        left: -7px;
```
planctus, line 61: This is kind of random css, i'm not even sure what this is supposed to do, but having this values -3px 7px -7px looks to me like something we would not really want to do.

src/themes/default/src/scss/_multiselect.scss
```diff
@@ -12,3 +12,60 @@
...
+    }
+    .ss-multi-selected {
+      background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 12 12' width='12' height='12' fill='none' stroke='%23dc3545'%3e%3ccircle cx='6' cy='6' r='4.5'/%3e%3cpath stroke-linejoin='round' d='M5.8 3.6h.4L6 6.5z'/%3e%3ccircle cx='6' cy='8.2' r='.6' fill='%23dc3545' stroke='none'/%3e%3c/svg%3e");
+      background-position: right 1.8rem center, center right 2.25rem;
+      background-size: 1.4rem 1.4rem;
+      background-repeat: no-repeat;
+      border-color: #dc3545;
+    }
+  }
+  .badge {
```
planctus, line 29: So we pretend this is a badge by altering the element via js, but i'm not really sure what we want to achieve by doing this...we set the primary color background by setting another class, but we could have it simply passing the right value for the $primary-color that slimselect.scss uses...why, then...? we have also border-radius available as a variable, what is the advantage of defining this as a badge. i wonder...

planctus (changes requested):
I believe we are not going anywhere with this, we are spending so much time around these small design nuances and to me there are things we could have done and we could do in a straightforward way and some others that is just not worth doing. But we choose to go in ways that would just result in a weak, clumsy and hard to maintain approach for both of them...

planctus:
Can you check the reason why of the difference in terms of font family or font style i see between the badge and these elements from slimselect..? I think the x as we have it is acceptable, the main difference i notice with the badge is indeed the font rendering, see if we can align that and then for me this lightweight solution can be good enough.

planctus (approved):

PR #43, opened 2021-08-10
feat(ISA): Create JoinUp navbar.

theosuja:
`variables.scss` file contains only the part of @gingapa 's work that is required for the navbar component to be built. Once this branch is merged into `master`, just keep his variables in case of conflict.


src/themes/joinup/src/scss/oe-bcl-joinup.scss
```diff
@@ -4,6 +4,9 @@
 @import "@openeuropa/bcl-bootstrap/scss/mixins";
 @import "@openeuropa/bcl-bootstrap/scss/utilities";
 
+// Override Bootstrap variables
+@import "base/variables";
```
planctus, line 8: Please import the scss files as in all the other cases from the npm package, the joinup theme is `@openeuropa/bcl-theme-joinup`

src/themes/joinup/src/scss/_navbar.scss
```diff
@@ -0,0 +1,472 @@
...
+        }
+        &::after {
+          width: 0;
+          height: 0;
+          margin-left: 0;
+          background: none;
+          transition: none;
+        }
+      }
+    }
```
planctus, line 470: Well, there is so much code for just this "component" i believe if we start like this this "theme" is soon going to be unmaintanable. The themes we provide should be bootstrap themes, i don't know if you could have saved any or many lines of code, but the feeling i have is that this is going to explode in our hands if we have to write tons of css code...

planctus (changes requested):

PR #45, opened 2021-08-12
feat(search-improvements): Improvements on search - FRONT-2785

tibi2303:


.storybook/preview-head.html
```diff
@@ -59,6 +59,7 @@
       }
 
       var offcanvasMobile = document.querySelectorAll(".offcanvas-mobile")[0];
+      var offcanvasMobileBody = document.querySelectorAll(".offcanvas-mobile .offcanvas-body")[0];
```
planctus, line 62: considering the fact that we have a different preview-head.html for each theme, and at the moment we don't have a way to unify them, i was wondering if it would not be smarter to set the javascript calls for the components using it directly in the stories, so that we could get rid of this duplication.
But here it seems to me that again, we are doing something improperly, it's true that this is a composition and not a component but if it needs javascript to work other than just the initialization of the offcanvas i think this js should be distributed by the library, bundled in the theme js file.

planctus, line 62: but so we want to get the first available selector and we use querySelectorAll()[0] instead of querySelector()..? :)

planctus (commented):
I don't know, i feel we do things just for the sake of a demo we make in our storybook instances, but i'm not really confident this is enough, we are dealing with a library of components that then defines also "layouts" using those components, ideally if we wouldn't have any single line of js for doing this it would be easier, but if rwe eally need js to make things work we have to start thinking about delivering it, bundling it with the library.

tibi2303:
I agree with having an external js to have this scripts inside it, but how we will deliver it? do you have any suggestions ? @planctus

tibi2303:
@planctus In order to overwrite the visibility hidden in inline style on offcanvas, I needed to add important in scss and disable the stylelint no-declaration-important on that file.

planctus (commented):
Greaaaaaat! :) 
So one single media query and a few lines of css and we've got rid of custom js, brilliant!
The only thing, which is not directly related to this pr but more to the "dismissible" badge..we render a close button, but it doesn't do anything..
Now, i would like to use a "bootstrap" way, but i think you have already tried the close-button idea and you had issues in styling that, maybe...
If that is not really possible i think we have to do it dirty, like with an onClick but i think if we define something as dismissible it has to be dismissible, otherwise the functionality is not there..

planctus (approved):

PR #46, opened 2021-08-16
feat(form-validation): Adding validation parameters to the form-input fields and demo improvement on forms - FRONT-2808

tibi2303:
After the multiselect pr is accepted and merged, in this pr needs to be added the multiselect success state.


src/components/bcl-form-input/form-input.story.js
```diff
@@ -80,6 +81,9 @@ const resetAttrs = (data, args) => {
   if (!args.invalid) {
     data.attributes.removeClass("is-invalid");
   }
+  if (!args.valid) {
```
planctus, line 84: You should add it also to the getArgs() function, without a default value we get the "set boolean" button.

planctus (changes requested):
The requested change is a small one, but to be honest, the implementation matches the ticket "add success state" but i think we could have been doing more, handling the form validation the way bootstrap does, i mean adding parameters also for the valid and invalid feedback and including the "form" component we could have also easily added examples of client validation in its stories. 
Anyway for the sake of the "ticket", only the control needs to have a default value set, the rest could be done afterwards if we like the idea...

.storybook/preview-head.html
```diff
@@ -8,6 +8,24 @@
   document.addEventListener(
     "DOMContentLoaded",
     function () {
+      // Example starter JavaScript for disabling form submissions if there are invalid fields
```
planctus, line 11: 1. 

1) Ok, but if we only add this js to this storybook we are the only one who are going to see this possibility...
I know we are not happy with the multiplication of the preview-head.html files and the duplication involved by that, but we don't have a solution for the moment, so we should add this in all the storybook instances.
Otherwise, as i was suggesting you in another pr, this js for a specific "component" could be only set in the story as a decorator, this way we would avoid the duplication in the preview-head files.
This has the consequence of appearing in the html tab, but i think is actually good so that we show the "example" js used to initialise something that is not available otherwise.
You can add a comment on top of it saying "Example code to initialize client side validation on the form"

2) I see that you applied this to the "basic" example, i would add an information like: `Basic (with client side validation)`, then, in the related story.

3) I see specs files with empty values for these newly added parameters, if you don't want to set them simply don't add them in the specs, but i think it would be actually better to pass some value, just to see them in action for a "server side" validation adding the classes through the controls for valid and invalid.

planctus (changes requested):
OK, now we have the parameters we were missing, but we cannot really see an example of the client side validation in amy of the published styleguides.
I think we can start moving the js code we use to initialize "optional" components/functionalities in the stories, if you agree you can do it for this example of client validation.

planctus:
Better, but:
i think it's confusing now, client side validation is always enabled but we advertise it only for the basic example where we apply the "bootstrap" custom validation, so...
1) i would apply the browser default only to one example, you specify in the basic that is bootstrap form validation and in this other example that is the browser default, instead.
2) we apply the `novalidate` attribute to the other examples so that it doesn't validate.
3) in those examples without client validation i would add an action to the form  like` onsubmit="return false;`", so that the submit button can still be used with no arm.
4) i expected the script to be included  in the html tab while is not, that's a pity..

planctus (approved):

PR #49, opened 2021-08-20
feat(user-listing): Added user listing - Front 2451

tibi2303:


src/components/bcl-card/card.html.twig
```diff
@@ -113,11 +142,17 @@
...
 {% endif %}
 {% if _card_header is not empty %}
-  <div class='card-header'>
+  <div class='{{ _header_classes }}'>
     {{- _card_header -}}
   </div>
 {% endif %}
-  <div class='card-body'>
+  <div class='{{ _body_classes }}'>
+  {% if _category is not empty %}
```
planctus, line 150: i'm not sure about this, i don't know if this would work, but ideally i would prefer these badges in the card-header more than in the body..

src/components/bcl-card/card.html.twig
```diff
@@ -145,9 +180,22 @@
   {% if _content is not empty %}
     {{ _content }}
   {% endif %}
+  {% if _card_tags is not empty and _card_tags is iterable %}
```
planctus, line 183: mmh..this is not so convincing...in such a scenario wouldn't this simply be the "content" of the card, why adding a new parameter for this..? I see that maybe the expectation is to have two fields to map to these tags thingy, but still this needs "work" on the implementation side to pass the object as we needed it..so this looks questionable to me.
Also the name is confusing, initially i thought that the "tags" where the badges, but you named that parameter "category", instead...

src/compositions/bcl-user-listing/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@openeuropa/bcl-user-listing",
+  "private": true,
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "0.5.0",
```
planctus, line 6: Well, we have just released 0.6.0, but you need to include this package in the dependencies of bcl-twig-components, otherwise the templates will not be available in any of the themes.

planctus (changes requested):
I think you should check the preview before considering a pull request ready for review, anyway it was pretty straighforward to find the issue, the implementation is pretty ok, i'm fine with the added parameters for the additional classes of the main elements of the card.
But i have a js error:
`Uncaught TypeError: Cannot read property 'addEventListener' of undefined`
which i think is causing the offcanvas not being opened by clicking on the trigger button in mobile.

src/components/bcl-card/card.html.twig
tibi2303: I moved it to header, it works but i needed to add some reset classes for the head default styles.

src/components/bcl-card/card.html.twig
tibi2303: Indeed, I removed this and added all in content.

src/compositions/bcl-user-listing/package.json
tibi2303: Updated everything to 0.6 and add user-listing package as a dependency in twig templates, I don't really like the naming because when we will have user-view etc we will have multiple package with almost the same name. It would be great to have one package 'user' and that to include everything regarding user packages as dependency but i think should be done when we will have also the view.

src/components/bcl-card/card.html.twig
```diff
@@ -113,11 +142,17 @@
...
-  <div class='card-header'>
+  <div class='{{ _header_classes }}'>
     {{- _card_header -}}
   </div>
 {% endif %}
-  <div class='card-body'>
+  <div class='{{ _body_classes }}'>
+  {% if _category is not empty %}
+    {% set _category = _category|merge({
+      attributes: create_attribute().addClass("mb-3")
```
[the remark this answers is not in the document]
tibi2303: Indeed, i added the wrapper.

src/components/bcl-card-layout/__snapshots__/card-layout.test.js.snap
```diff
@@ -88,7 +88,7 @@ exports[`OE - Card grid renders correctly 1`] = `
           </p>
           
         Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
-    
+
```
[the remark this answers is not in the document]
tibi2303: I removed the white spaces and empty lines.

src/compositions/bcl-user-listing/user-listing.story.js
```diff
@@ -0,0 +1,35 @@
+import demoData from "@openeuropa/bcl-user-listing/dataUserListing.js";
+import userListing from "@openeuropa/bcl-user-listing/user-listing.html.twig";
+import defaultSprite from "@openeuropa/bcl-bootstrap/bootstrap-icons.svg";
+
+export default {
+  title: "Compositions/Users/Listing",
+};
+
+const scriptInit = (story) => {
```
planctus, line 9: mmh..i thought you forgot to add this, but it's not working in the demo...

src/compositions/bcl-user-listing/user-listing.story.js
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const correctPaths = (data) => {
+  data.data.filter_button.icon.path = defaultSprite;
+  data.data.badges.forEach((badge) => {
+    badge.icons_path = defaultSprite;
+  });
+  return data;
+};
+
```
planctus, line 33: it's a minor thing, but i don't see the reason why we use "placeholder" as the selected option of the select elements, i think it would be better to try to imagine real use cases and create demos with realistic content, but i understand is an additional overhead (also the "organisation position" thingy in the teasers could be improved). But in any case use one of the existing options, or none, don't add a "placeholder" that is not a placeholder

src/compositions/bcl-user-listing/user-listing.story.js
```diff
@@ -0,0 +1,35 @@
...
+export const Default = () => userListing(correctPaths(demoData));
```
planctus, line 34: I think we miss something in the css, i see that you didn't touch it here, but testing a bit the element i noticed that when expanding the viewport from mobile when the offcanvas is visible something bad happens. (i don't know why but i'm unable to attach a screenshot, you should be able to reproduce this easily, though)

src/compositions/bcl-user-listing/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@openeuropa/bcl-user-listing",
+  "private": true,
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "0.6.0",
+  "description": "OE - BCL search",
```
planctus, line 7: Update this description text..

src/components/bcl-card/card.html.twig
```diff
@@ -41,6 +41,10 @@
     options: ['dark', 'white']
   - card_header (string) (default: '')
   - card_footer (string) (default: '')
+  - extra_classes_body (string) (default: '')
+  - extra_classes_header (string) (default: '')
+  - extra_classes_footer (string) (default: '')
+  - category (string) (default: {} - badge object)
```
planctus, line 47: this is an array of badge objects, right..? and surely not a string.. :)

planctus (changes requested):
It seems pretty much ok to me, please check the comments one by one, but it seems we are close to get this..

src/compositions/bcl-user-listing/package.json
planctus: I understand, but for the way the subtheme will potentially override these templates it would not work if we have multiple templates in the same package, i'm actually thinking that at the moment it would not be possible to override the offcanvas elements in these two compositions. But yeah, for the moment there is no need.

src/compositions/bcl-user-listing/user-listing.story.js
tibi2303: I added the fix now, but when the offcanvas is opened it's also adding on the body overflow hidden and a padding right.
On this case, it's ok if i override it on the mobile viewport, but i'm thinking of the offcanvas default component which will be impacted by this (we will have the scrollbar when offcanvas is opened on desktop).

src/components/bcl-card/card.html.twig
```diff
@@ -112,42 +135,47 @@
...
 <div class='col-{{ _horizontal_grid.right_col_responsive ? _horizontal_grid.right_col_responsive ~ '-' }}{{ _horizontal_grid.right_col }}'>
 {% endif %}
-{% if _card_header is not empty %}
-  <div class='card-header'>
+{% if _card_header is not empty or _category is not empty %}
+  <div class='{{ _header_classes }}'>
     {{- _card_header -}}
+    {%- if _category is not empty -%}
+      <div class="mb-3">
+        {% include '@oe-bcl/bcl-badge/badge.html.twig' with _category only %}
```
planctus, line 143: Ah wait, i haven't noticed this before, i assumed this was supporting an array of "badges", i mean, why would we limit the "category" to one..? Once we add this let's support multiple badges and maybe, "badges" would also be the right label for this new parameter...

src/compositions/bcl-user-listing/user-listing.html.twig
```diff
@@ -0,0 +1,45 @@
...
+<main>
+  <div class="container mt-5">
+    <div class="row">
+      <div class="col-md-3">
+        {% include '@oe-bcl/bcl-user-listing/user-listing-offcanvas.html.twig' with data.offcanvas only %}
+      </div>
+      <div class="col-12 col-lg-9">
+      {% include '@oe-bcl/bcl-button/button.html.twig' with data.filter_button only %}
+      <div class="float-end d-none d-md-block">
+        {% include '@oe-bcl/bcl-select/select.html.twig' with data.sort_select|merge({
```
planctus, line 15: not so relevant considering that this is a composition, but still since the themes will ship these templates i would still like to keep them clean, why don't we add these classes in the story (if you don't want to duplicate the elements in the specs)..?

planctus (approved):

PR #54, opened 2021-09-06
feat(user-edit): Added composition of user edit - Front 3040

tibi2303:


planctus:
you forgot to add this new package as a dependency in bcl-twig-templates...
In any case, once we merge this, we will put everything in one single package, the limitation we had before is not anymore in place, so no reason why multiplicating the packages (even if they are private ones)

planctus (approved):

PR #55, opened 2021-09-06
chore: Removing ECL, using ecl header as default, updating compositions

planctus:


src/compositions/bcl-header/header.story.js
```diff
@@ -0,0 +1,24 @@
+import { withDesign } from "storybook-addon-designs";
+import headerECLData from "@openeuropa/bcl-data-header/data-ecl";
+import header from "@openeuropa/bcl-header/header.html.twig";
+import "!!null-loader!@openeuropa/bcl-theme-default/src/scss/_header.scss";
```
tibi2303, line 4: do we need this? we are adding it in base scss for each theme i see.

planctus: You mean the "import" of the scss..? This is a fake import, it's the way to have the scss code in the panel on storybook, but that !!null means that is not really imported.. ;)

tibi2303 (approved):

PR #58, opened 2021-09-09
feat(Update-scripts): Moved script inits - FRONT-3129

tibi2303:


src/components/bcl-scrollspy/scrollspy.story.js
```diff
@@ -22,6 +45,8 @@ export default {
 
 export const Navigation = () => navScrollspy;
 Navigation.storyName = "Navigation";
+Navigation.decorators = [initScrollspy];
```
planctus, line 48: but in a such a scenario, where we apply the decorator to each story, wouldn't it be better to set the decorator in the export default directly...?

planctus (commented):
Great! Just i think in a few story instead of setting the decorator in each export you can simply do it in the export default, but it works fine on all the components, we don't have the hassle to maintain this html files anymore, too good ;)

src/components/bcl-scrollspy/scrollspy.story.js
tibi2303: I moved the decorators to export story for offcanvas, scrollspy and toasts.

PR #60, opened 2021-09-14
feat(footer): Added composition of footer - Front 3064

tibi2303:
We are waiting for links-block to be integrated here.


src/themes/default/src/scss/_footer.scss
```diff
@@ -0,0 +1,3 @@
+.oe-footer {
+  background: $gray-100;
```
tibi2303, line 2: We will need to sync regarding this variables.. it's not the same background as on figma, also the border color it's different

src/compositions/bcl-footer/footer.story.js
```diff
@@ -0,0 +1,41 @@
...
+
+const correctPaths = (data) => {
+  data.rows.forEach((row) => {
+    row.cols.forEach((col) => {
+      if (col.items) {
+        col.items.forEach((item) => {
+          if (item.type == "links") {
+            item.links.forEach((link) => {
+              if (link.icon) {
+                link.icon.path = defaultSprite;
```
tibi2303, line 28: I don't really like this but it may change after we have social content paragraph in place.

src/compositions/bcl-footer/footer.html.twig
```diff
@@ -0,0 +1,83 @@
...
+                  },
+                  ...
+                ],
+              },
+            ]
+          }
+        ]
+      },
+      ...
+    ]
```
planctus, line 36: wow, try to reduce this comment.. :)

src/compositions/bcl-footer/footer.html.twig
```diff
@@ -0,0 +1,83 @@
...
+#}
+
+{% set _rows = rows|default([]) %}
+
+<footer class="oe-footer">
```
planctus, line 41: i know it's not "final", but let's assume that these templates we make for compositions might be used as such, use drupal attrs on its root element.

src/compositions/bcl-footer/footer.story.js
```diff
@@ -0,0 +1,41 @@
...
+            });
+          }
+        });
+      }
+    });
+  });
+  return data;
+};
+
+export const Footer = () => footer(correctPaths(footerData));
```
planctus, line 39: Really a minor thing, but why not exporting this as `Default` so that you don't have to set its storyName..? :)

tibi2303:
The differences between this implementation and links block has some big difference, i decided to keep it as is.

src/compositions/bcl-footer/footer.html.twig
```diff
@@ -0,0 +1,84 @@
...
+                        class="{{ item.classes }}"
+                      {% endif %}
+                    >
+                  {% endif %}
+                  {% if item.type == 'content' %}
+                    {{ item.content }}
+                  {% endif %}
+                  {% if item.type == 'links' %}
+                    {% for link in item.links %}
+                      {% include '@oe-bcl/bcl-link/link.html.twig' with link|merge({
```
papegaill, line 69: Not so obvious to validate the spacing next to figma, we must sync with the designers for that. but here I will add an `mb-1` class to give more space between the links (see figma)

src/compositions/bcl-footer/footer.story.js
papegaill: is this ok?

src/data/footer/data.js
```diff
@@ -0,0 +1,237 @@
...
+            },
+            {
+              type: "links",
+              links: [
+                {
+                  label: "Facebook",
+                  icon_position: "before",
+                  icon: {
+                    name: "facebook",
+                    size: "xs",
```
papegaill, line 45: on these social icons links, I'll put a `me-2` extra_class to give more horizontal space as on figma

papegaill (changes requested):
Here are some suggestions for spacings.
I also see EC blue variant on figma, is this requested?

src/data/footer/data.js
```diff
@@ -0,0 +1,237 @@
...
+            },
+          ],
+        },
+        {
+          classes: "col-12 col-lg-4 pb-4 pb-lg-0",
+          items: [
+            {
+              type: "content",
+              content:
+                "<p class='fw-bold border-bottom pb-2'>Contact [project name]</p>",
```
papegaill, line 22: not sure, but all the titles seems to have a margin-bottom higher than specified in figma.
maybe a `mb-2` to follow design?

papegaill (changes requested):

papegaill (approved):

PR #61, opened 2021-09-14
feat(links block): Add composition of links block - Front 3136

theosuja:


src/compositions/bcl-links-block/dataLinksBlock.js
```diff
@@ -0,0 +1,21 @@
+const drupalAttribute = require("drupal-attribute");
+
+module.exports = {
+  data: {
```
tibi2303, line 4: you don't need to add all specs inside a data object, just keep it in root

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,29 @@
+{% spaceless %}
+
+{# Parameters:
+  - title
+  - links
+    - path
+    - label
+#}
+
+{% set _title = data.title|default('') %}
```
tibi2303, line 10: keep it simple, only title not data.title. same for links

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,29 @@
+{% spaceless %}
+
+{# Parameters:
+  - title
```
tibi2303, line 4: more descriptive regarding this (you should use the template from one of the already built component)

src/compositions/bcl-links-block/links-block.story.js
```diff
@@ -0,0 +1,17 @@
...
+      {
+        name: "Wireframe",
+        type: "figma",
+        url: "https://www.figma.com/file/NQlGvTiTXZYN8TwY2Ur5EI/BCL-Features?node-id=1126%3A38732",
+      },
+    ],
+  },
+};
+
+export const LinksBlock = () => linksBlock(demoData);
```
tibi2303, line 17: add an end of line here, you should have a setting in visual studio that does this automatically on save

tibi2303: same goes for most of files

tibi2303:
Rename the pr title, to comply with the semantic check, just use one of the already merged pr template title.

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,29 @@
...
+    - path
+    - label
+#}
+
+{% set _title = data.title|default('') %}
+{% set _links = data.links|default([])%}
+
+<div class="bg-light p-3">
+  {% if _title is not empty %}
+    <p class="fs-5 fw-bold mb-2">{{ _title }}</p>
```
tibi2303, line 15: Here it should be a title object that has 3 props {label: "", classes: "", tag: ""}. You should be able to modify the classes and the tag if you want.

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,29 @@
...
+    <hr class="divider"/>
+  {% endif %}
+  {% if _links is not empty and _links is iterable %}
+    <ul class="ps-0 mb-0">
+      {% for _link in _links %}
+        <li class="list-unstyled pb-2">
```
tibi2303, line 21: you should add the margin on the link not on the li from my point of view.. add it using drupal_attributes object (addClass('mb-2'))

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,29 @@
...
+{% set _title = data.title|default('') %}
+{% set _links = data.links|default([])%}
+
+<div class="bg-light p-3">
+  {% if _title is not empty %}
+    <p class="fs-5 fw-bold mb-2">{{ _title }}</p>
+    <hr class="divider"/>
+  {% endif %}
+  {% if _links is not empty and _links is iterable %}
+    <ul class="ps-0 mb-0">
```
tibi2303, line 19: Here the static classes should come from the data spec not hardcoded

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,29 @@
...
+  - title
+  - links
+    - path
+    - label
+#}
+
+{% set _title = data.title|default('') %}
+{% set _links = data.links|default([])%}
+
+<div class="bg-light p-3">
```
tibi2303, line 13: Here the static classes should come from the data spec not hardcoded

tibi2303: just add {{ attributes }} and from there should be simple enough to add anything you need, not only classes.

src/compositions/bcl-links-block/links-block.html.twig
theosuja: I think these two particular selectors are quite "mandatory" since they override the <ul> reboot. I don't think it makes sense to make them configurable

src/compositions/bcl-links-block/links-block.html.twig
theosuja: For now the background isn't going to be optional, it may be in the future or in other compositions. I don't think the padding should be configurable though, any opinion @planctus @papegaill ?
For me it's not superclear yet how is this going to look like once implemented in Drupal, but I feel like this level of configuration is not necessary. Please let me know your thoughts about it.

src/compositions/bcl-links-block/links-block.html.twig
theosuja: 100% agreed

src/compositions/bcl-links-block/links-block.story.js
theosuja: thanks :)

src/compositions/bcl-links-block/links-block.html.twig
planctus: Well, i think this approach (including the link) is the same that is currently used in the template used for the "ECL" link-block, but at the same time we already have a bootstrap component handling list of links and i wonder whether the "OEL" design shouldn't be based on that, this is a bootstrap based library in the end..

.storybook/preview.js
```diff
@@ -54,7 +54,7 @@ export const parameters = {
   options: {
     storySort: {
       method: 'alphabetical',
-      order: ['Components', 'Compositions'],
+      order: ['Components', 'Compositions', 'Paragraphs'],
```
planctus, line 57: Nice that you "sorted" this Paragraphs thing, but then you placed the example in "compositions" :)

src/compositions/bcl-links-block/links-block.html.twig
planctus: Well, the approach we should have is that we offer the possibility via the template to reproduce the design we have, if then we know that one default "variant" has to be set we define this in the template as "default".
In this case it means that:
1) we are using drupal attributes here
2) we add a parameter for the background ( gray, white ), with a default value gray, that you use in to add this class otherwise you add another one for the white bg

padding i agree that might not be a parameter, but you will add it using attributes.addClass()

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,56 @@
...
+  {% set _title_attributes = create_attribute() %}
+{% endif %}
+
+{% set _classes = ['fs-5', 'fw-bold', 'mb-2'] %}
+
+{% if title.classes is not empty %}
+  {% set _classes = _classes|merge([title.classes]) %}
+{% endif %}
+
+{% if title.tag is not empty %}
```
planctus, line 28: do we actually need this..? I mean, if in the design we have one style only i don't see why we would give this possibility.
in any case i believe that "title" on drupal means just a string, so in case we need to maintain this configurable it would be better doing it as a different parameter, not making the title an object.

theosuja: The first version only had the title as a string, then Tiberiu kindly suggested me to add some options (and I kindly agreed :) ) but in this stage of my life I believe the less configurations are given, the better it is. So if @tibi2303 agrees, it's your decision @planctus

src/compositions/bcl-links-block/links-block.html.twig
planctus: talked with the designers, let's trick to this implementation that includes link, although i think this is a duplicate of an existing component, basically

src/compositions/bcl-links-block/links-block.story.js
```diff
@@ -0,0 +1,17 @@
+import demoData from "@openeuropa/bcl-links-block/dataLinksBlock.js";
+import linksBlock from "@openeuropa/bcl-links-block/links-block.html.twig";
+
+export default {
+  title: "Compositions/Links Block",
```
planctus, line 5: Put it in "Paragraphs"

.storybook/preview.js
theosuja: I thought you agreed with @tibi2303 to finally add it here. The commits show that we had it configured as a component in a new section "Paragraphs" but then we moved it back to "Compositions".
I'll ping you to sync on this

src/compositions/bcl-links-block/links-block.html.twig
tibi2303: sure, well i just proposed from the footer point of view, for me it would have been helpful to have this configurable, you can reverse it :D

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,56 @@
...
+  - links (array of objects) (default: []) format: [
+    path: (string) (default: '#')
+    label: (string) (default: '')
+  ]
+#}
+
+{% set _title = title|default({}) %}
+{% set _links = links|default([])%}
+
+{% if _title_attributes is empty %}
```
planctus, line 18: well, we never used title_attributes (drupal) for any component, this would be the first time, but what kind of extra params would we expect for the title..?
I think we don't need this either, basically..

src/compositions/bcl-links-block/links-block.story.js
planctus, line 17: Once you add the parameter to switch the background you will also have to:
a) create another export for the "other" background
b) add a control named "variant" to make the switch between the two.

Up to you ;)

src/compositions/bcl-links-block/links-block.html.twig
theosuja: Would it make sense to use a boolean since we only have 2 possibilities and a default?

src/paragraphs/bcl-links-block/links-block.test.js
```diff
@@ -0,0 +1,14 @@
+import { renderTwigFileAsNode } from "@openeuropa/bcl-test-utils";
+
+import demoData from "@openeuropa/bcl-links-block/dataLinksBlock";
+
+const template = "@oe-bcl/bcl-links-block/links-block.html.twig";
+const render = (params) => renderTwigFileAsNode(template, params);
+
+describe("OE - Links block", () => {
```
tibi2303, line 8: Can you add here a test for horizontal and grey bg? to have a full suite with all cases

tibi2303: Also did we agreed to move them back to paragraphs? Just asking, if it's the case, after the full test suite it's done, should be ok for me. @planctus if you feel the need to add more, please do :D

package.json
```diff
@@ -61,6 +61,7 @@
       "src/data/*",
       "src/components/*",
       "src/compositions/*",
+      "src/paragraphs/*",
```
planctus, line 64: No i see that we made confusion:
in storybook and storybook only we group this into a "paragraphs" labelled thing but there is no need to add another folder in the filesystem with the overhead that you saw, adding this folder here in the package.json, adding the support for the stories, etc..all these things can be reverted and the package should be in src/compositions

src/paragraphs/bcl-links-block/dataLinksBlock.js
```diff
@@ -0,0 +1,21 @@
...
+  variant: "gray",
+  direction: "vertical",
+  title: "More information",
+  links: [
+    {
+      path: "#",
+      label: "European Commission",
+    },
+    {
+      path: "#",
```
planctus, line 13: we link everything to /example.html, we have an example page for that ;)

src/paragraphs/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,69 @@
...
+{% set _variant = variant|default("gray") %}
+{% set _direction = direction|default("vertical") %}
+{% set _title = title|default("") %}
+{% set _links = links|default([])%}
+
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+{% if list_attributes is empty %}
```
planctus, line 24: i don't think we have to use drupal attributes here, i mean this is only an internal processing and you only set classes, isn't it easier to just use a string..?

theosuja: I am not 100% it's the only way to do it, but I did this to use the logic (l:52) to add classes depending on the variant chosen

theosuja: indeed the attribute creation can be simplified, but as far as I unerstand it can't be skipped

src/paragraphs/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,69 @@
...
+        {% if loop.last == false %}
+          {% set list_attributes = list_attributes.addClass(["me-4"]) %}
+        {% endif %}
+        <li 
+          {{ list_attributes }}
+        >
+          {% include "@oe-bcl/bcl-link/link.html.twig" with {
+            path: _link.path,
+            label: _link.label,
+            attributes: create_attribute().addClass(["pb-2", "d-inline-block"])
```
planctus, line 61: The spacing is pretty "approximated", i see that there are values like 20px in the mock and we cannot reproduce that with utilities unless we create additional spacing classes. But this in particular looks a bit evident, it's 16px in the mocks and we make it 8px..i think we should be a bit more precise and if we cannot reproduce exactly what we are asked to we should report that in the figma file and agree with the designers.

src/paragraphs/bcl-links-block/links-block.story.js
```diff
@@ -0,0 +1,63 @@
+import demoData from "@openeuropa/bcl-links-block/dataLinksBlock.js";
+import linksBlock from "@openeuropa/bcl-links-block/links-block.html.twig";
+
+const getArgs = (data) => {
+  return {
+    variant: data.variant || "gray",
```
planctus, line 6: you have only one data file and these two values are available, so why the || in these cases..?

src/paragraphs/bcl-links-block/links-block.story.js
```diff
@@ -0,0 +1,63 @@
...
+    direction: data.direction || "vertical",
+  };
+};
+
+const getArgTypes = (data) => {
+  return {
+    variant: {
+      type: { name: "select" },
```
planctus, line 14: if it's inline-radio why the type "select", you can remove this or you can replace its value with inline-radio, but simply as type: "inline-radio"

src/paragraphs/bcl-links-block/links-block.story.js
```diff
@@ -0,0 +1,63 @@
...
+      control: {
+        type: "inline-radio",
+      },
+      table: {
+        defaultValue: { summary: "gray" },
+        category: "Style",
+      },
+    },
+    direction: {
+      type: { name: "select" },
```
planctus, line 26: Also here

src/themes/default/storybook/bcl-builder.config.js
```diff
@@ -5,7 +5,7 @@ const outputFolder = path.resolve(__dirname);
 module.exports = {
   copy: [
     {
-      from: ["../../../{components,compositions}/**/*.story.js"],
+      from: ["../../../{components,compositions,paragraphs}/**/*.story.js"],
```
planctus, line 8: so this will go away..

src/themes/joinup/storybook/bcl-builder.config.js
```diff
@@ -8,15 +8,15 @@ const excludePaths = [];
 if (stories.length) {
   stories.forEach((story) => {
     excludePaths.push(
-      `../../../{components,compositions}/**/${story}.story.js`
+      `../../../{components,compositions,paragraphs}/**/${story}.story.js`
```
planctus, line 11: this also..

src/themes/joinup/storybook/bcl-builder.config.js
```diff
@@ -8,15 +8,15 @@ const excludePaths = [];
...
     );
   });
 }
 
 module.exports = {
   copy: [
     {
-      from: ["../../../{components,compositions}/**/*.story.js"],
+      from: ["../../../{components,compositions,paragraphs}/**/*.story.js"],
```
planctus, line 19: same..

theosuja: yes! those changes are ready to ship

src/paragraphs/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,69 @@
+{% spaceless %}
+
+{# Parameters:
+  - variant: (string) (default: "gray")
+    - options ["gray", "white"]
+  - direction: (string) (default: "vertical")
+    - options ["vertical", "horizontal"]
```
planctus, line 7: but you define white as an accepted value for this parameter but you don't use it, you should set the `bg-white` class if white is the value provided.

theosuja: indeed

.storybook/environment.js
```diff
@@ -8,6 +8,7 @@ const drupalAttribute = require("drupal-attribute");
 
 const projComponentsAbsPath = path.resolve(__dirname, "../src/components");
 const projCompositionsAbsPath = path.resolve(__dirname, "../src/compositions");
+const projParagraphsAbsPath = path.resolve(__dirname, "../src/paragraphs");
```
planctus, line 11: this will disappear

.storybook/environment.js
```diff
@@ -16,6 +17,7 @@ if (typeof loader.addPath === "function") {
   // Add namespace oe.
   loader.addPath(projComponentsAbsPath, "oe-bcl");
   loader.addPath(projCompositionsAbsPath, "oe-bcl");
+  loader.addPath(projParagraphsAbsPath, "oe-bcl");
```
planctus, line 20: and then this..

.storybook/main.js
```diff
@@ -1,6 +1,6 @@
 const path = require("path");
 
-const stories = ["../src/{components,compositions}/**/*.story.js"];
+const stories = ["../src/{components,compositions,paragraphs}/**/*.story.js"];
```
planctus, line 3: this line can be reverted as well

planctus (changes requested):
A few small thing, it's kind of ok but the "design" compliance is not really high, there are many things not perfectly reproducing the mocks.

src/paragraphs/bcl-links-block/links-block.story.js
theosuja: I must have copied this from another component where it was used

src/paragraphs/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,69 @@
...
+
+{% set attributes = attributes.addClass(_classes) %}
+{% set list_attributes = list_attributes.addClass(_list_classes) %}
+
+<div 
+  {{ attributes }}
+>
+  {% if _title is not empty %}
+    <h5 class="fs-5 fw-bold mb-2">{{ _title }}</h5>
+    <hr/>
```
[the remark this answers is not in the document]
theosuja: that's right

src/paragraphs/bcl-links-block/links-block.test.js
theosuja: thanks @tibi2303 for the support

src/paragraphs/bcl-links-block/links-block.html.twig
theosuja: Now I doubt if it has to be white or transparent actually... i'll ask the designers

theosuja:
> A few small thing, it's kind of ok but the "design" compliance is not really high, there are many things not perfectly reproducing the mocks.

Pointed the "impossible" values to the designers, once they fix those I'll make sure we are aligned and submit the PR again

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,63 @@
...
+  {% set _list_classes = _list_classes|merge(["d-inline"]) %}
+{% endif %}
+
+{% set attributes = attributes.addClass(_classes) %}
+
+<div 
+  {{ attributes }}
+>
+  {% if _title is not empty %}
+    <h5 class="fs-5 fw-bold pb-2 border-bottom">{{ _title }}</h5>
```
planctus, line 41: Thanks for removing the hr, but what i see here is that you add a 0.5 rem padding bottom (pb-2) and it should be pb-3

planctus, line 41: and, since you are dealing with the designers about the spacing inconsistencies, would you also ask if the h5 is the right heading from their point of view (mainly a semantical concern)...

theosuja: image
I assumed that one, but it's better to ask indeed...

theosuja: It's 12px on the mocks, I had to choose between 8 and 16 so I arbitrarily chose 8. Do you arbitrarily decide that it must be 16?

planctus: true, forget it, i thought we had 0.75rem, these spacing utilities are pretty limiting, i think we should extend them in order to have all the 4px steps...

theosuja: As I told this morning to Julien, I completely agree. We should have some more flexibility on this very specific utility. Who has the authority to decide on this? :)

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,63 @@
...
+    <h5 class="fw-bold pb-3 mb-3 border-bottom">{{ _title }}</h5>
+  {% endif %}
+  {% if _links is not empty and _links is iterable %}
+    <ul class="ps-0 mb-0">
+      {% for _link in _links %}
+        {% if loop.last == false %}
+          {% set _list_classes = _list_classes|merge(["me-4-5"]) %}
+        {% endif %}
+        <li 
+          {{ create_attribute().addClass(_list_classes) }}
```
planctus, line 50: i think there is really zero value in using drupal attributes here, you can simply work on a string and print it as it is, here we create an object to only store a few classes name with no interaction with drupal, we use drupal attributes for the root of the component and the included templates, so everywhere drupal might be adding his own attrs, not on internal elements, basically...

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,63 @@
...
+        >
+          {% include "@oe-bcl/bcl-link/link.html.twig" with {
+            path: _link.path,
+            label: _link.label,
+            attributes: create_attribute().addClass(["pb-3", "d-inline-block"])
```
planctus, line 55: what you should here, instead, is to check whether any "attributes" exists already in the data, if not you create a new one otherwise you just add your classes.

src/themes/default/src/scss/_spacers.scss
```diff
@@ -0,0 +1,11 @@
+// scss-docs-start spacer-variables-maps
```
planctus, line 1: ok, cool, but Tiberiu is adding a _variables.scss, let's not multiply them, rename this in _variables.scss, we will then merge also the variables from the Tiberiu pr...

src/compositions/bcl-links-block/__snapshots__/links-block.test.js.snap
```diff
@@ -0,0 +1,145 @@
...
+      </li>
+    </ul>
+  </div>
+</jest>
+`;
+
+exports[`OE - Links block renders correctly with transparent background 1`] = `
+<jest>
+  <div
+    class=""
```
planctus, line 102: why do we have this empty attribute..?

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,63 @@
...
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+{% set _classes = [] %}
+{% set _list_classes = ["list-unstyled"] %}
+
+{% if _variant == "gray" %}
+  {% set _classes = _classes|merge(["bg-light", "px-4", "py-3"]) %}
+{% endif %}
```
planctus, line 29: can you please ensure that the variant is supposed to be transparent and not white..? Maybe that is the case but since a few things depend on this let's just be sure about it.

src/compositions/bcl-links-block/links-block.story.js
```diff
@@ -0,0 +1,61 @@
...
+        category: "Style",
+      },
+    },
+    direction: {
+      description: "Direction in which the links are displayed",
+      options: ["vertical", "horizontal"],
+      control: {
+        type: "inline-radio",
+      },
+      table: {
```
planctus, line 30: you should add also a "type" "string" so that it will appear in the "table"

src/compositions/bcl-links-block/links-block.story.js
```diff
@@ -0,0 +1,61 @@
...
+        defaultValue: { summary: "gray" },
+        category: "Style",
+      },
+    },
+    direction: {
+      description: "Direction in which the links are displayed",
```
planctus, line 25: you should add also a "type" "string" so that it will appear in the "table"

src/compositions/bcl-links-block/links-block.html.twig
theosuja: fair enough

src/themes/default/src/scss/_spacers.scss
planctus: please remove also the docs comment, that's from bootstrap..

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -0,0 +1,54 @@
...
+        {% if _link.attributes is empty %}
+          {% set _link = _link|merge({
+            attributes: create_attribute()
+          }) %}
+        {% endif %}
+        {% set _link = _link|merge({
+          clean_class: true,
+          attributes: _link.attributes.addClass(["pb-3", "d-inline-block"])
+        }) %}
+        <li class="list-unstyled{% if _direction == "horizontal" %} d-inline{% endif %}{% if loop.last == false %} me-4-5{% endif %}">
```
planctus, line 46: i would prefer a cleaner, although more verbose way of setting these classes.
```
{% _list_item_classes = "list_unstyled" %}
{% if _direction == "horizontal" %}
{% set _list_item_classes = _list_item_classes ~ ' d-inline' %}
{% endif %}
{% if loop.last %}
{% set _list_item_classes = _list_item_classes ~ '  me-4-5' %}
{% endif %}
```

PR #63, opened 2021-09-16
feat(header): Added header compositions - Front 3065

tibi2303:


src/themes/default/src/scss/_header.scss
```diff
@@ -1,38 +1,116 @@
...
 .oe-header__navbar {
-  background-color: $header-background;
+  background: linear-gradient(
+    90deg,
+    $header-background-start 0%,
+    $header-background-end 100%
+  );
+}
+
+.ecl__header .oe-header__project,
```
planctus, line 21: I think i'm confused by this naming, i know that we used to have a "default" and an "ecl" like version, that's why ecl was mentioned but now we have two different versions, one for EC and one for EU, both "ecl" based, so i think the rereference to ECL should completely disappear and it can be replaced by EC/EU

src/themes/default/src/scss/_header.scss
```diff
@@ -1,38 +1,116 @@
...
 $header-margin: 1.5rem 0.5rem !default;
 
 .oe-header {
-  display: flex;
-  align-items: center;
-  justify-content: space-between;
   margin: $header-margin;
 }
 
+.oe-header__project,
```
planctus, line 12: Also oe is outdated, now we are bcl so if we have to write any new markup this should be prefixed with our "namespace", bcl and not oe anymore

src/compositions/bcl-header/header.html.twig
```diff
@@ -1,78 +1,79 @@
...
+      {% if _project_logo.image.alt is not empty %}
+        alt="{{ _project_logo.image.alt }}"
+      {% endif %}
+      {% if _project_logo.image.classes is not empty %}
+        class="{{ _project_logo.image.classes }}"
+      {% endif %}
+      />
+    {% endif %}
+    {% if _project_logo.title is not empty %}
+      <p>{{ _project_logo.title }}</p>
```
planctus, line 56: And we would wrap the site name in a <p>..? :)
I believe it should be an heading, maybe an <h1>..

src/compositions/bcl-header/header.html.twig
```diff
@@ -1,78 +1,79 @@
...
-    - modal
-  - language
-    - link
-      - label
-      - href
-      - target
-    - modal
+  - navbar (Navbar object) (default: {})
+  - project_logo (object) (default: {})
+    format: {
```
planctus, line 6: mmh..well, we know that this might just be a "composition", but at the same time i wouldn't nest the site name into a "logo" kind of parameter, i think you can have `site_name` as one the parameters of the template and separately the logo

src/compositions/bcl-header/header.html.twig
```diff
@@ -1,78 +1,79 @@
...
+        image: {
+          path: "",
+          alt: "",
+          classes: ""
+        },
+        title: ""
+      }
+  - navbar_settings (Navbar object) (default: {})
```
planctus, line 14: mmh..i don't really like this parameter name, wouldn't it be better like simply `topbar` or `head`?

src/themes/default/src/scss/_header.scss
planctus: Design wise, i've seen in the mocks this area is white for the EU version, not sure you have more up to date info about this

src/themes/default/src/scss/_header.scss
```diff
@@ -1,38 +1,116 @@
...
+.oe-header__navbar .navbar-nav .nav-link.active {
   color: #fff;
 }
 
+.oe-header .notification {
+  position: relative;
+}
+
+.oe-header .navbar-brand img {
+  width: 100px;
```
planctus, line 50: this 100px is maybe fine for the mobile, but already in tablet it looks very small, maybe that is the breakpoint where it should change size (i know we miss a tablet version in the design, we should probably raise this concern)

src/themes/default/src/scss/_header.scss
```diff
@@ -1,38 +1,116 @@
...
+}
+
+.oe-header .navbar-toggler {
+  position: absolute;
+  right: 14px;
+  bottom: calc(100% - 0.45rem);
+  border-color: $gray-500;
+}
+
 @media (min-width: 992px) {
```
planctus, line 88: shouldn't this be: @include media-breakpoint-up(lg)?

src/themes/default/src/scss/_header.scss
```diff
@@ -1,38 +1,116 @@
...
 
+.oe-header .notification {
+  position: relative;
+}
+
+.oe-header .navbar-brand img {
+  width: 100px;
+}
+
+.oe-header .notification span {
```
planctus, line 53: we need to improve the styles of these links, the color the alignment with the icons, but good that you have found a reasonable way of adding this notification icon.. ;)

planctus (changes requested):
See the comments, but it seems pretty okish..theo is adding some spacing utilities maybe you will be able to use them to fine tune some of the spacings here..

src/compositions/bcl-header/header.html.twig
tibi2303: that's how was i added at first but i saw in figma they changed it from heading... not sure, i will ask romain

tibi2303: Romain told me that this should be a p or a div, i will let it as a p.

src/themes/default/src/scss/_header.scss
tibi2303: this improvement will come when we will have the new color variables and spacing.

papegaill (changes requested):
Really nice !! just a few remarks.
We should also see for the login/language modals, no design for this yet?

src/data/header/data-ec.js
```diff
@@ -0,0 +1,248 @@
...
+      items: [
+        [
+          {
+            type: "text",
+            required: true,
+            label: "Search",
+            hidden_label: true,
+            placeholder: "Search",
+            id: "inlineFormInputGroupSearch",
+            attributes: new drupalAttribute().addClass([
```
papegaill, line 106: possible to have the same light focus outline as the button for the search input? we don't see the blue outline on a blue bg: 

Screenshot 2021-09-23 at 10 39 07

src/themes/default/src/scss/_header.scss
```diff
@@ -1,38 +1,132 @@
...
+  width: 100px;
+}
+
+.bcl-header .notification span {
+  position: absolute;
+  top: 18px;
+  right: 8px;
+  width: 18px;
+  height: 18px;
+  padding-left: 6px;
```
papegaill, line 62: I'll go for a padding-left of 5px; the number is not correctly center(mac - chrome)
6px:
Screenshot 2021-09-23 at 10 42 22
 
5px:
Screenshot 2021-09-23 at 10 42 00
 
Also what if we have more than 1 digit (badge should be extensible):
Screenshot 2021-09-23 at 10 44 31

papegaill (changes requested):

src/data/header/data-ec.js
papegaill: details.. but to me the search icon is not correctly center in the button, she seems to be leaning down :)
I will put a margin-top of -2px on the icon's svg, up to you..

actually:
Screenshot 2021-09-23 at 11 12 41

with negative top margin: 
Screenshot 2021-09-23 at 11 12 27

src/themes/default/src/scss/_header.scss
```diff
@@ -1,38 +1,141 @@
...
+}
+
+.bcl-header__project.light p {
+  color: $black;
+}
+
+.bcl-header .navbar-toggler {
+  position: absolute;
+  right: 14px;
+  bottom: calc(100% - 0.45rem);
```
papegaill, line 92: `bottom: calc(100% - 0.3rem);` to correctly center the toggle vertically?

src/themes/default/src/scss/_header.scss
```diff
@@ -1,38 +1,141 @@
...
+  width: 100px;
+}
+
+.bcl-header .notification span {
+  position: absolute;
+  top: 18px;
+  right: 9px;
+  height: 18px;
+  padding-right: 5px;
+  padding-left: 5px;
```
papegaill, line 67: thank you for the change, but I think the border radius and the position is confusing with 2 digit:

Screenshot 2021-09-23 at 15 15 49

By removing absolute position and use transform, and change border to 10px, here is a better result, I think:

2 digit:
Screenshot 2021-09-23 at 15 16 42

1 digit:
Screenshot 2021-09-23 at 15 16 56


```
border-radius: 10px;
height: 18px;
line-height: .6rem;
padding-left: 5px;
padding-right: 5px;
transform: translate(-11px, 5px);
z-index: 1;
```

papegaill (changes requested):

papegaill (approved):

PR #65, opened 2021-09-17
fix(card): Header and footer, improving stories - FRONT-3165

planctus:


tibi2303:
I know this is the intended markup in bootstrap, but it look a bit weird when we have a header card and the image as a background, it overlays.

tibi2303:
Overall is good, i will approve it if nothing can be regarding the previous comment.

planctus:
You're right, the example was looking bad, i added a reset of the header and footer when you choose "background" as the image position.

tibi2303 (approved):

PR #67, opened 2021-09-20
feat(variables): Added new colors. - FRONT-3180

tibi2303:


src/themes/default/src/scss/_variables.scss
```diff
@@ -0,0 +1,7 @@
+// Colors variables
+
+$blue-50: #cfe2ff;
```
planctus, line 3: I've spent a bit of time trying to understand whether we shouldn't merge these values in the existing maps and not only defining them as variables, but i'm not sure it would make any difference..

planctus (approved):

PR #68, opened 2021-09-20
feat(tooltip): Adding control for tooltip position in button and link - FRONT-3172

planctus:


src/components/bcl-button/button.story.js
```diff
@@ -121,6 +127,21 @@ const getArgTypes = (data) => {
...
   };
+  if (type === "tooltip") {
+    argTypes.tooltip_position = {
+      name: "tooltip position",
+      type: "select",
+      options: ["top", "bottom", "left", "right"],
+      description: "Position for the tooltip",
+      table: {
+        type: { summary: "attribute" },
+        defaultValue: { summary: "top" },
```
tibi2303, line 138: We don't have a default value for the tooltip, this is set only through drupal attributes, i'm not sure but it may be a bit misleading.

src/components/bcl-button/button.story.js
```diff
@@ -179,7 +204,7 @@ const initTooltip = (story) => {
         return new bootstrap.Tooltip(tooltipTriggerEl);
       });
     </script>
-  ${demo}`;
+    <div style="padding: 2rem 0 2rem 8rem" class="bg-light">${demo}</div>`;
```
tibi2303, line 207: It's only a personal opinion but i don't think we need the bg-light.

src/components/bcl-button/button.story.js
tibi2303: Or i any case, i think it should be bottom, right? You set it earlier

planctus: no, this is bootstrap default. https://getbootstrap.com/docs/5.0/components/tooltips/#markup

src/components/bcl-button/button.story.js
planctus: didn't like it either, i removed it ;)

tibi2303 (approved):

PR #69, opened 2021-09-20
feat(alert): Adding icons to the alert component - FRONT-3176

planctus:


src/components/bcl-alert/alert.html.twig
```diff
@@ -6,14 +6,27 @@
...
 
 {% set _message = message|default('') %}
 {% set _variant = variant|default('primary') %}
 {% set _heading = heading|default('') %}
+{% set _icons_path = icons_path|default('') %}
 {% set _dismissible = dismissible ?? true %}
 {% set _animated_dismiss = animated_dismiss ?? true %}
-{% set _classes = ['alert', 'alert-' ~ _variant|e('html_attr')] %}
+{% set _classes = ['alert', 'alert-' ~ _variant|e('html_attr'), 'd-flex', 'align-items-center'] %}
+{% set _icon_names = {
```
tibi2303, line 19: It's a nice approach but in the future, if they will change the icons file it might not have the same names (this might also happen if bs will update the icons), we should not limit this. We can do this in the story to have a nice control for the demo.

src/components/bcl-alert/alert.html.twig
```diff
@@ -25,10 +38,22 @@
 {% set attributes = attributes.addClass(_classes).setAttribute('role', 'alert') %}
 
 <div {{ attributes }}>
-{%- if _heading is not empty -%}
-  <h4 class="alert-heading">{{ _heading }}</h4>
-{%- endif -%}
-  {{- _message -}}
+{% if _icons_path is not empty %}
```
tibi2303, line 41: If we change this to have it without icons_name, I think this can be done only checking for icon, if it exists.

src/data/alert/data.js
```diff
@@ -3,5 +3,6 @@ const drupalAttribute = require("drupal-attribute");
 module.exports = {
   message: "A simple alert. check it out!",
   heading: "Alert heading",
+  icons_path: "/icons.svg",
```
tibi2303, line 6: I always set this in the story to the defaultSprite. I'm not sure if should be done the same here.

src/components/bcl-alert/alert.html.twig
planctus: mmh..no, i think that the association of icons with alert variants is something chosen by the "designer", it's not a free choice by the user and it seems to me it's the template that can make this association without any additional parameter.

src/components/bcl-alert/alert.html.twig
planctus: Well, i don't want an icon_name param, it's not needed here

src/data/alert/data.js
planctus: yeah, in the story we set it to the "real" path, these are the specs and i think we normally set this to /icons.svg

tibi2303 (approved):

PR #70, opened 2021-09-21
feat(multiselect): Bug fix and large size. - FRONT-3178

tibi2303:


papegaill (commented):
We have small and large sizes, but what about normal size(medium). 
Same as the form input, it seems that this default size is missing, 
or maybe a control issue?

papegaill (changes requested):
Icon display issue for the large version when no option within the field:
Screenshot 2021-09-27 at 12 11 45
Screenshot 2021-09-27 at 12 11 37
To me icon can be a bit smaller ;)

papegaill:
> @papegaill I actually found a problem here, we don't actually have the medium size, the normal size of the multiselect is the sm size of the select, so we need to change the default multiselect size to something bigger, and keep the default as the small size.

Thks, can we avoid using `form-control-md` class as it doesn't seems to be part of boostrap? 
https://getbootstrap.com/docs/4.0/components/forms/#sizing
Can we keep same approach as all forms elements?

I also noticed previously that all the form elements on the search page and maybe other ones are on small sizes, not sure this is correct, maybe we will have to change everything to default(medium) sizes, to confirm.. if yes, will ba another ticket.

tibi2303:
@papegaill thanks for the remarks, we need to confirm the sizes from search and user pages with the designers, anyway i removed the md size. 👍

src/components/bcl-select/select.html.twig
```diff
@@ -37,12 +37,14 @@
 {% set _label_class = 'form-label' %}
 {% set _hidden_label = hidden_label|default(false) %}
 
-{% set _classes = _classes|merge(['form-select-' ~ _size]) %}
-
 {% if _clean_class %}
   {% set _classes = [] %}
 {% endif %}
 
+{% if _size is not empty %}
```
planctus, line 44: mmh..i believe what we are doing here is going to conflict with what we will need later...the behavior of these form elements in terms of their size is not anymore left to the user choosing a size for it and it should actually be handled via css media queries basing on the viewport.
What has been said is that the size would be "lg" in mobile viewport and "md" in the others as far as i remember

src/themes/default/src/scss/_multiselect.scss
```diff
@@ -26,4 +27,83 @@
...
+  }
+  .ss-values {
+    .ss-value {
+      font-size: 1rem;
+    }
+    .ss-disabled {
+      padding: 8px 5px;
+    }
+  }
+  &.form-select-sm {
```
planctus, line 51: basically this adaptations should not be based on the class but on a media query, if i'm correct

planctus (commented):
I believe the "styles" are good, but the approach needs probably to be changed basing on this:
https://citnet.tech.ec.europa.eu/CITnet/confluence/display/FRONT/BCL+updates+to+align+on+OEL

src/components/bcl-select/select.html.twig
tibi2303: done. hope it's good

src/themes/default/src/scss/_multiselect.scss
tibi2303: done.

planctus:
I'm confused, think we do too many things not in the right order, in any case since we are dealing with these sizes for all the form elements i think you can already remove everything related to the size from the template and the story

tibi2303:
@planctus I agree, the good order should be : 
1. https://github.com/openeuropa/bootstrap-component-library/pull/81 (sizes) - I already remove the sizes you said in this pr.
2. This pr
3. Search form

planctus (approved):

PR #71, opened 2021-09-21
feat(social media follow): Add social media follow paragraph - Front 3140

theosuja:


src/compositions/bcl-links-block/dataLinksBlock.js
```diff
@@ -1,8 +1,10 @@
 const drupalAttribute = require("drupal-attribute");
 
 module.exports = {
-  variant: "gray",
-  direction: "vertical",
+  extra_classes: "",
```
tibi2303, line 4: you can delete this if you don't set, also for extra_attributes

src/compositions/bcl-social-media-follow/dataSocialMediaFollow.js
```diff
@@ -0,0 +1,79 @@
+const drupalAttribute = require("drupal-attribute");
+
+module.exports = {
+  extra_classes: "",
```
tibi2303, line 4: delete this if you don't set them, also for extra_attributes

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -12,16 +19,28 @@
...
+{% set _variant = variant|default("vertical") %}
 {% set _title = title|default("") %}
 {% set _links = links|default([])%}
 
 {% if attributes is empty %}
   {% set attributes = create_attribute() %}
 {% endif %}
 
-{% if _variant == "gray" %}
+{% if _extra_classes is not empty %}
```
tibi2303, line 33: actually, you don't need extra_classes nor extra_attributes, you can set them from specs directly, delete them

tibi2303:
I'm not sure but we don't need a separate twig file for social media, you can use links-block and all that data that you set in social media twig can come from the specs directly, it's overcomplicated.

src/compositions/bcl-links-block/__snapshots__/links-block.test.js.snap
```diff
@@ -72,7 +72,7 @@ exports[`OE - Links block renders correctly with horizontal 1`] = `
         </a>
       </li>
       <li
-        class="list-unstyled d-inline  me-4-5"
+        class="list-unstyled  me-4-5"
```
tibi2303, line 75: There are 2 spaces between the classes, can we fix this?

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -1,27 +1,39 @@
...
+  - extra_attributes: (array of objects) (default: []) format: [
+    {
+      "name" (string) (default: ''),
+      "value" (optional) (string),
+    }
+  ]
+  - background: (string) (default: "gray")
     - options ["gray", "transparent"]
-  - direction: (string) (default: "vertical")
+  - variant: (string) (default: "vertical")
```
tibi2303, line 13: I don't really like the naming, maybe we can have a boolean param here called horizontal, default false and if it's set to true should have the behaviour of horizontal... it's fine tuning, but i think it would be better.

src/compositions/bcl-links-block/links-block.story.js
```diff
@@ -38,6 +40,12 @@ const getArgTypes = (data) => {
 };
 
 const applyArgs = (data, args) => {
+  for (let i = 0; i < data.links.length; i++) {
```
tibi2303, line 43: make it a forEach here

src/compositions/bcl-links-block/links-block.story.js
```diff
@@ -61,3 +69,9 @@ export const LinksBlock = (args) => linksBlock(applyArgs(demoData, args));
 
 LinksBlock.args = getArgs(demoData);
 LinksBlock.argTypes = getArgTypes(demoData);
+
+export const SocialMediaFollow = (args) =>
+  linksBlock(applyArgs(socialMediaFollowData, args));
```
tibi2303, line 74: I don't know if it's because of prettier but make it in one line here.

src/compositions/bcl-links-block/__snapshots__/links-block.test.js.snap
theosuja: sure !

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -1,27 +1,39 @@
 {% spaceless %}
 
 {# Parameters:
-  - variant: (string) (default: "gray")
+  - extra_classes: (string) (default: "")
```
tibi2303, line 4: forgot to delete this. extra_classes extra_attributes

src/compositions/bcl-links-block/links-block.html.twig
theosuja: it is like this to match the previous paragraph... it was explicitly asked to be like this as far as possible

theosuja: "match the previous paragraph" means that this variable is currently called "variant" and the options are horizontal and vertical. It seems that it facilitates the migration to keep those names when possible

src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -1,27 +1,39 @@
...
+      "value" (optional) (string),
+    }
+  ]
+  - background: (string) (default: "gray")
     - options ["gray", "transparent"]
-  - direction: (string) (default: "vertical")
+  - variant: (string) (default: "vertical")
     - options ["vertical", "horizontal"]
   - title: (string) (default: "")
   - links (array of objects) (default: []) format: [
```
tibi2303, line 16: here can be only array of link object and it's enough

src/compositions/bcl-links-block/links-block.story.js
theosuja: it is indeed because of prettier, i'll try but maybe it will be automatically changed :)

src/compositions/bcl-links-block/links-block.html.twig
tibi2303: you don't need to add the format after

src/compositions/bcl-links-block/links-block.html.twig
tibi2303: ah ok, didn't checked that, thanks

theosuja: @papegaill do you confirm it's better to keep the former variables naming?

papegaill (approved):

PR #72, opened 2021-09-23
feat(search): Added search - FRONT-3177

tibi2303:


tibi2303:
I tried to adapt as much as i could to the default form input behaviour, the only problem i have is regarding the floating label, i tried many possible solutions but none works, if you have any idea please.

papegaill:
> I tried to adapt as much as i could to the default form input behaviour, the only problem i have is regarding the floating label, i tried many possible solutions but none works, if you have any idea please.

for me we can remove the control for the floating label, I don't see the interest for this component to have this option.

src/components/bcl-form-input/form-input.html.twig
```diff
@@ -148,10 +153,17 @@
 {% endif %}
 {% if _horizontal and floating == false %}
   <div class="{{ _horizontal_class }}">
+{% endif %}
+{% if _icon is not empty %}
+<div class="{{ _icon_container_classes }}">
+  {%- include '@oe-bcl/bcl-icon/icon.html.twig' with _icon only -%}
 {% endif %}
     <input
       {{ attributes }}
```
papegaill, line 162: input type need to be `search` instead of `text`

papegaill: to see if we can style the webkit search cancel button: https://stackoverflow.com/questions/22928819/override-the-webkit-search-cancel-button

papegaill (changes requested):
Nothing happen with the required control and as mentioned, to me we can remove the floating option/control ;)

src/components/bcl-form-input/form-input.html.twig
papegaill: this should do the trick for styling the search cancel button ;)

```
.bcl-search-bar__input[type="search"]::-webkit-search-cancel-button {
  -webkit-appearance: none;
  height: 16px;
  width: 16px;
  background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 16 16' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M16 8C16 12.4183 12.4183 16 8 16C3.58172 16 0 12.4183 0 8C0 3.58172 3.58172 0 8 0C12.4183 0 16 3.58172 16 8ZM5.35355 4.64645C5.15829 4.45118 4.84171 4.45118 4.64645 4.64645C4.45118 4.84171 4.45118 5.15829 4.64645 5.35355L7.29289 8L4.64645 10.6464C4.45118 10.8417 4.45118 11.1583 4.64645 11.3536C4.84171 11.5488 5.15829 11.5488 5.35355 11.3536L8 8.70711L10.6464 11.3536C10.8417 11.5488 11.1583 11.5488 11.3536 11.3536C11.5488 11.1583 11.5488 10.8417 11.3536 10.6464L8.70711 8L11.3536 5.35355C11.5488 5.15829 11.5488 4.84171 11.3536 4.64645C11.1583 4.45118 10.8417 4.45118 10.6464 4.64645L8 7.29289L5.35355 4.64645Z' fill='%236C757D'/%3E%3C/svg%3E");
  cursor: pointer;
}
```

tibi2303: I don't have safari on windows, i pushed the improvements, thanks. Can you please check it if it's working properly?

papegaill: the type for the input is still `text`

tibi2303: done.

src/components/bcl-form-input/form-input.html.twig
```diff
@@ -21,6 +21,7 @@
   - "hidden_label" (boolean) (default: false)
   - "size" (string) (default: '')
   - "switch" (boolean) (default: false)
+  - "icon" (icon object) (default: {})
```
planctus, line 24: mmh..i don't think this is the way, the icon can be added in the composition i believe, and the form-input can stay as it is,  this is both because this is the only use case we have with an icon and because we also need an additional wrapper, we are making a composition template and that is where these elements should be added.

src/components/bcl-icon/icon.html.twig
```diff
@@ -13,6 +13,7 @@
 {% set _path = path|default('') %}
 {% set _size = size|default('s') %}
 {% set _transformation = transformation|default('') %}
+{% set _wrapper_classes = wrapper_classes|default('') %}
```
planctus, line 16: what is this supposed to be used for..?

src/compositions/bcl-search-bar/search-bar.html.twig
```diff
@@ -0,0 +1,20 @@
+{% spaceless %}
+
```
planctus, line 2: It seems to me that for how you made it this could have been done simply with the form template, but if we move here instead the addition of the icon and the wrapper it might make a bit more sense as a composition.

src/compositions/bcl-search-bar/search-bar.story.js
```diff
@@ -0,0 +1,89 @@
+import demoData from "@openeuropa/bcl-search-bar/dataSearch.js";
+import searchBar from "@openeuropa/bcl-search-bar/search-bar.html.twig";
+import { getFormControls } from "@openeuropa/bcl-story-utils";
+import defaultSprite from "@openeuropa/bcl-bootstrap/bootstrap-icons.svg";
+import drupalAttribute from "drupal-attribute";
+
+const getArgs = (data) => {
```
planctus, line 7: We miss the size in the controls and i think it's important

planctus (changes requested):
In the mocks i see in the second part of the page a different design, i'm not sure what we are supposed to do with that, you probably have more information about this..

src/components/bcl-form-input/form-input.html.twig
tibi2303: indeed, but at the markup level of the input comes a lot of elements (label, valid/invalid messasge, helper), i need the icon and only the input inside a container because i set the icon with absolute. I know it adds complexity with what we don't really need but i can't find any other way to be able to have all the functionality of the form input that has an icon.

tibi2303:
From what i know, the data list is not needed now, we don't have enough info.

src/components/bcl-icon/icon.html.twig
tibi2303: Forgot about it, done.

src/compositions/bcl-search-bar/search-bar.story.js
tibi2303: added for search also.

src/components/bcl-form-input/form-input.html.twig
tibi2303: done.

planctus:
In mobile i see this: 
image

src/compositions/bcl-search-bar/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@openeuropa/bcl-search-bar",
+  "private": true,
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "0.7.0",
+  "description": "OE - BCL search bar",
+  "dependencies": {
+    "@openeuropa/bcl-form-input": "^0.7.0"
```
planctus, line 9: now we need to add also the icon in the dependencies

src/compositions/bcl-search-bar/search-bar.html.twig
```diff
@@ -0,0 +1,24 @@
...
+{% set _pill = pill|default(false) %}
+{% set _icon = icon|default({}) %}
+
+{% set classes = "bcl-search-bar" %}
+
+{% if _pill %}
+  {% set classes = classes ~ " rounded" %}
+{% endif %}
+
```
planctus, line 17: i would add also a wrapper where you would render `attributes` but actually i would use also the bcl-search-bar class on it.
In ECL we have a label (for accessibility) but we hide it, i wonder if we shouldn't do the same also here.
Furthermore, shouldn't this composition actually render a "form" tag and it could then maybe renamed as search-form instead of search-bar..?

planctus:
in the preview i see that the size control is not working properly but it will disappear, so maybe we can just remove it

src/compositions/bcl-search-bar/search-bar.html.twig
tibi2303: The input label already does that, we have it display: none in css, right now you won't have a label displayed on this input.

I added the input_attributes to not overlap with the attributes from the form but still be able to use controls for the story.

The icon is shown correctly on mobile, when we will merge the sizes it will looks good, that's intended.

I also removed the size control for search type but still it will be removed entirely in the sizes pr.

src/compositions/bcl-search-form/search-form.story.js
```diff
@@ -0,0 +1,31 @@
...
+
+const correctPaths = (data) => {
+  data.icon.path = defaultSprite;
+
+  return data;
+};
+
+export const Default = () => searchForm(correctPaths(demoData));
+
+Default.storyName = "Default";
```
planctus, line 27: You need to set the storyName only in case its string is different than the name of the exported story. In these two cases it's not needed.

tibi2303: done

src/themes/default/src/scss/_forms.scss
```diff
@@ -0,0 +1,46 @@
+.bcl-search-form {
+  .bcl-search-form__group {
+    position: relative;
+  }
+  .bcl-search-form__icon {
+    position: absolute;
+    top: 17px;
```
planctus, line 7: it looks to me like an "empirical" way of writing css, like this.
The size of the select in mobile is 3rem, the size of the icon is 1rem, so it's value should be (select-size - icon-size) / 2 which is indeed 1rem, 16px vertical, don't get the 17px value, mobile is 11px because (38-16) / 2 makes 11px, but it should be a computed value, not us "hardcoding" it.
Let's try to use rems and not pixels and always use the $spacers map when it's about spacing, more than "hardcoded" values.

src/compositions/bcl-search-form/search-form.html.twig
```diff
@@ -0,0 +1,31 @@
...
+
+{% set _classes = ["bcl-search-form"] %}
+
+{% if _pill %}
+  {% set _classes = _classes|merge(['rounded']) %}
+{% endif %}
+
+{% set attributes = attributes.addClass( _classes ) %}
+
+<form
```
planctus, line 22: We should also close this tag at some point.. :)

src/compositions/bcl-search-form/search-form.html.twig
```diff
@@ -0,0 +1,31 @@
...
+  {{ attributes }}
+>
+  <div class="bcl-search-form__group">
+    {% include '@oe-bcl/bcl-icon/icon.html.twig' with _icon only %}
+    {% include '@oe-bcl/bcl-form-input/form-input.html.twig' with _input only %}
+  </div>
+</div>
```
planctus: but..we don't have a submit button in this form..? :) How is this search going to work..?

src/compositions/bcl-search-form/search-form.html.twig
tibi2303: done ...

src/compositions/bcl-search-form/search-form.html.twig
tibi2303: but that's the way it's done in figma, i'm not sure, that's why i added in the beginning as a form input.

planctus: what do you mean..? in figma there is the way it should look like, then the way it "works" is a matter of the implementation, to me that icon is actually the submit button, indeed.

tibi2303: @planctus i refactored the stories and add the mission one. Should be good now. Sorry for all the ping pong on this one.

planctus (approved):

PR #74, opened 2021-09-27
feat(user-small-teaser): Create composition for small user view - FRONT-3175

theosuja:
I have serious doubts if it is worth it to use the card component for this. As I understand, it must be edited to handle one new configuration: row/col classes different than the classic layout ones (not "row" and "col-XX", but others). Mainly because we need the image to have a fixed width/height, not responsive
 
I think the layer of complexity added to handle this simple configuration is not worth it, I believe it would be better to create a new component which would be quite simple and specific.

Please let me know what you think @planctus @tibi2303 

If we decide to continue in this way, the card must be refactored (at least the way we handle horizontal variant and the row/col classes) and the test must be updated. And probably the story so we display this new simple version of the card


src/components/bcl-card/card.html.twig
```diff
@@ -25,6 +25,7 @@
       path (string),
       alt (string),
       position (string)
+      size_auto (boolean)
```
planctus, line 28: Are all these changes a leftover since when you tried to use a card..?

src/compositions/bcl-media-object/media-object.html.twig
```diff
@@ -0,0 +1,51 @@
...
+      - "alt" (string) (default: '')
+      - "extra_classes" (optional) (string) (default: '')
+    }
+  - text (object) format: {
+      - "content" (string) (default: '')
+      - "extra_classes" (optional) (string) (default: '')
+    }
+#}
+
+{% if attributes is empty %}
```
planctus, line 15: mmh..well, i don't necessarily see why shipping this as a "media-object" template, it is true that it might turn to be useful, but i wouldn't consider this "view mode" for the user profile an implementation of a media-object "component" so for me this should be in the "user" series of compositions, bcl-user-something..

src/compositions/bcl-media-object/media-object.html.twig
```diff
@@ -0,0 +1,51 @@
...
+  {% set attributes = create_attribute() %}
```
planctus, line 16: attributes exists for all the templates in drupal, we never check for their "presence".

src/compositions/bcl-media-object/media-object.html.twig
```diff
@@ -0,0 +1,51 @@
...
+{% endif %}
+
+{% if media.extra_classes is not empty %}
+  {% set _media_attributes = create_attribute() %}
```
planctus, line 20: but why do we have to create a drupal attributes object, can't we simply add those classes coming from the data instead to the class attribute..?

src/compositions/bcl-media-object/media-object.html.twig
```diff
@@ -0,0 +1,51 @@
...
+  {% set _media_attributes = _media_attributes.addClass(media.extra_classes) %}
+{% endif %}
+
+{% set _text_attributes = create_attribute() %}
```
planctus, line 24: same here, we use drupal attributes only for the root element and the included templates, because there might be attributes coming from the specs, it's not convenient for us otherwise especially in a simple case like we have here

src/compositions/bcl-media-object/media-object.html.twig
```diff
@@ -0,0 +1,51 @@
...
+  {{ attributes }}
+>
+  <div class="flex-shrink-0">
+    <img
+      src="{{ media.path }}"
+      alt="{{ media.alt }}"
+      {{ _media_attributes }}
+    >
+  </div>
+  <div
```
planctus, line 44: i think this classes should be hardcoded here, we are implementing a design, creating a composition, we shouldn't require extra data to be passed, only the "content".

src/compositions/bcl-media-object/media-object.html.twig
```diff
@@ -0,0 +1,51 @@
+{% spaceless %}
+
+{# Parameters:
+  - media (object) format: {
+      - "path" (string) (default: '')
+      - "alt" (string) (default: '')
+      - "extra_classes" (optional) (string) (default: '')
+    }
+  - text (object) format: {
```
planctus, line 9: but why nesting this into an object, i think this could be flattened

src/compositions/bcl-media-object/media-object.html.twig
```diff
@@ -0,0 +1,51 @@
+{% spaceless %}
+
+{# Parameters:
+  - media (object) format: {
```
planctus, line 4: i believe that once renamed in a bcl-user-whatever this parameter could also be renamed to "picture"

src/compositions/bcl-media-object/media-object.html.twig
```diff
@@ -0,0 +1,51 @@
...
+      - "path" (string) (default: '')
+      - "alt" (string) (default: '')
+      - "extra_classes" (optional) (string) (default: '')
+    }
+  - text (object) format: {
+      - "content" (string) (default: '')
+      - "extra_classes" (optional) (string) (default: '')
```
planctus, line 11: In any case it seems pretty poor in terms of a data structure for this composition, everything goes into a single key of this text object, couldn't we do something better having a "title" or "user_name" and separately an array of items like, i don't know, "roles" where you hardcode in the template the classes you need instead of expecting them from the specs..?

src/compositions/bcl-media-object/media-object.html.twig
```diff
@@ -0,0 +1,51 @@
...
+{% endif %}
+
+{% set attributes = attributes.addClass(["d-flex"]) %}
+
+<div
+  {{ attributes }}
+>
+  <div class="flex-shrink-0">
+    <img
+      src="{{ media.path }}"
```
planctus, line 39: this is really surprising, we first create a drupal attribute but then we hardcode in the html of the template two attributes..? As i told you here drupal attributes don't make sense so i'm fine with the attributes declared directly in the html, remove the creation of the the drupal attrs and pass also the class as you are doinf with these.

src/compositions/bcl-media-object/media-object.story.js
```diff
@@ -0,0 +1,31 @@
+import demoData from "@openeuropa/bcl-media-object/dataMediaObject.js";
+import demoDataSmallUser from "@openeuropa/bcl-media-object/dataSmallUser.js";
```
planctus, line 2: i don't know why we started going camelCase for these data in the compositions, it's not more than a loose convention, but we normally name these specs data-something-else

src/compositions/bcl-media-object/media-object.story.js
```diff
@@ -0,0 +1,31 @@
...
+      {
+        name: "Bootstrap docs",
+        type: "iframe",
+        url: "https://getbootstrap.com/docs/5.0/utilities/flex/#media-object",
+      },
+    ],
+  },
+};
+
+export const SmallUserView = (args) =>
```
planctus, line 27: I think we only need this demo here, but can you set a human readable name for it..?

src/compositions/bcl-media-object/media-object.story.js
```diff
@@ -0,0 +1,31 @@
+import demoData from "@openeuropa/bcl-media-object/dataMediaObject.js";
+import demoDataSmallUser from "@openeuropa/bcl-media-object/dataSmallUser.js";
+import mediaObject from "@openeuropa/bcl-media-object/media-object.html.twig";
+
+const applyArgs = (data, args) => {
+  return Object.assign(data, args);
+};
+
+export default {
+  title: "Compositions/Media object",
```
planctus, line 10: This will go in the user "folder" in storybook, ok..?

src/compositions/bcl-user-listing/user-listing.html.twig
```diff
@@ -19,7 +19,7 @@
             extra_classes: "w-auto d-inline-block ms-2"
           }) only %}
         </div>
-        <h4 class="mb-4">Users (18)</h4>
+        <h4 class="mb-4">Users ({{ data.cards.length }})</h4>
```
planctus, line 22: and how is this related to this work..?

planctus (changes requested):

planctus:
Ah, i don't think this is the expected result in case of "multiple" information, i believe that the image would stay top aligned, can you confirm with the designers..? Also the layout of the "roles" or other information is not clear to me, it might just be as it is right now, one info after the other, but please ensure that instead is not a "column" based layout..

src/components/bcl-card/card.html.twig
theosuja: oh crap, yes they are

src/compositions/bcl-media-object/media-object.story.js
theosuja: Not sure... you mean the SmallUserView? Anyway it's not relevant anymore

src/compositions/bcl-media-object/media-object.story.js
theosuja: I could be wrong, but I think this is inherited from other components and compositions. I'll try to avoid it

src/compositions/bcl-user-listing/user-listing-teaser.html.twig
```diff
@@ -0,0 +1,47 @@
+{% spaceless %}
```
planctus, line 1: as said previously i don't think we need this

planctus (changes requested):

src/compositions/bcl-user-listing/user-teaser.html.twig
```diff
@@ -0,0 +1,39 @@
...
+{# Parameters:
+  - picture (object) format: {
+    - "path" (string) (default: '')
+    - "alt" (string) (default: '')
+  }
+  - name (object) format: {
+    - "text" (string) (default: '')
+    - "link" (string) (default: '')
+  }
+  - department (string) (default: '')
```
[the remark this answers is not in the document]
theosuja: indeed, just forgot to update the doc.....

src/compositions/bcl-user-view/user-view-compact.html.twig
```diff
@@ -0,0 +1,37 @@
...
+>
+  <div class="flex-shrink-0">
+    <img
+      class="rounded-pill"
+      src="{{ picture.path }}"
+      alt="{{ picture.alt }}"
+    >
+  </div>
+  <div class="flex-grow-1 ms-3">
+    <a href="{{ name.link }}" class="fs-5 text-decoration-none mb-0">{{ name.text }}</a>
```
planctus, line 28: sorry to point this out only now, but normally when we have a title we don't assume that it has to be a link, i bet that in this case these will be always links but still let's put some logic to alternatively render simply the "text" if the "link" is empty.

src/compositions/bcl-user-view/user-view-compact.story.js
```diff
@@ -0,0 +1,20 @@
+import demoDataCompact from "@openeuropa/bcl-user-view/dataUserViewCompact.js";
```
planctus, line 1: there is no reason why you had to create a different story file, add this story to the existing one.

src/compositions/bcl-user-view/user-view-compact.story.js
```diff
@@ -0,0 +1,20 @@
...
+import userViewCompact from "@openeuropa/bcl-user-view/user-view-compact.html.twig";
+
+export default {
+  title: "Compositions/Users/View",
```
planctus, line 5: This is also explaining the reason why you don't have padding around your component, the default story of user-view is setting the layout to fullscreen, you are duplicating here the same "title" so you basically inherit that setting. When you will move it into the user-view.story.js you will have to override in the Compact.parameters object.

planctus (changes requested):

planctus (approved):

planctus (approved):

PR #76, opened 2021-09-28
feat(joinup): Fine-tune btn-light - FRONT-3210

theosuja:


src/themes/joinup/src/scss/_buttons.scss
```diff
@@ -24,3 +24,10 @@
     color: $brand;
   }
 }
+
+.btn-light {
```
planctus, line 28: not so important, but since the same exact properties are set to the secondary variant i wonder why not adding this new selector to the existing one.

planctus (approved):

PR #78, opened 2021-09-28
fix(multiselect): Add invalid state during validation - FRONT-3226

papegaill:
see existing slimselect issue: 
https://github.com/brianvoe/slim-select/issues/81

a screencast with the workaround is available on the ticket:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FRONT-3226?focusedCommentId=4963802&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4963802

### DIFF
--- a/src/components/bcl-select/select.html.twig
+++ b/src/components/bcl-select/select.html.twig
@@ -69,6 +69,11 @@
 
 {% if _required %}
   {% set attributes = attributes.setAttribute('required', true) %}
+
+  {% if _multiple %}
+    {% set attributes = attributes.setAttribute('oninvalid', "this.slim.slim.container.classList.add('is-invalid')") %}
+    {% set attributes = attributes.setAttribute('onchange', "this.slim.slim.container.classList.remove('is-invalid')") %}
+  {% endif %}
 {% endif %}
 
 {% if _hidden_label %}
--- a/src/components/bcl-select/__snapshots__/select.test.js.snap
+++ b/src/components/bcl-select/__snapshots__/select.test.js.snap
@@ -62,6 +62,8 @@ exports[`OE - select renders correctly as a multiple choice select 1`] = `
   <select
     class=""
     multiple="true"
+    onchange="this.slim.slim.container.classList.remove('is-invalid')"
+    oninvalid="this.slim.slim.container.classList.add('is-invalid')"
     required="true"
   >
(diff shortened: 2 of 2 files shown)


src/components/bcl-select/select.html.twig
```diff
@@ -69,6 +69,11 @@
 
 {% if _required %}
   {% set attributes = attributes.setAttribute('required', true) %}
+
+  {% if _multiple %}
```
planctus, line 73: mmhh...i don't think we should do this...basically we would introduce an exception on the client side validation behaviour by doing it, none of the form input in a form with client side validation gets the "invalid" style while a "pop up" is showed next to the field (in chrome at least).
With this we do the other way around, the pop up is not shown and the field is styled as invalid.

I think that we should simply "hide" the select in a different way, most likely with the:
```
  position: absolute;
  clip: rect(0, 0, 0, 0);
  pointer-events: none;  
```
- This would make the client validation (browser default) work as for the other input fields.

- Then we need to consider the bootstrap validation, which adds a .was-validated class to the form, we can use this to style the slimselect as the invalid one since it's just a sibling of the .form-select:invalid selector.

- Finally we have the server side validation for which we already use the is-invalid class and that is fine.

papegaill: Thank you for the feedback, I also thought of this solution which is indeed more realistic. I just didn't use the `clip` property who became obsolete.

papegaill: Thanks for the bootstrap validation, I added the selectors for the `:isinvalid` state during the validation, but no way to do the same for `:valid` state. it seems impossible to select sibling in a parent container `.was-validated`.

planctus: Two things:

- the `display: inline-block` we are using to override the `display: none` set by slimselect needs `!important` to be effective.

- Then, it seems to me that you applied the "bootstrap" styles to the standard browser validation, by doing this the multiselect would be the only form element having this look and feel when using this validation, i think those styles should only be used when the form has the "was-validated" class (which means bootstrap validated the form)

papegaill: Thank you 👍 everything should be ok now!

tibi2303 (approved):

PR #81, opened 2021-09-30
feat(Sizes): Redo form elements sizes. - FRONT-3229

tibi2303:


planctus (commented):
Good! but we broke a bit the styles of the switch, there is no margin anymore between the "widget" and the label, also the big size in mobile looks maybe a bit weird...

planctus (approved):

PR #84, opened 2021-10-05
feat(multiselect): Improvements - FRONT-3250

tibi2303:


planctus (changes requested):
Mmhh..what is going wrong with these styles..?
image
I see now we have again a huge font-size in desktop and i don't see any change in the badge changing the viewport, basically..

PR #117, opened 2021-11-03
feat(compositions): Add Timeline composition - FRONT-2571

papegaill:
Preview: 
- https://preview-117--oelibrary.netlify.app/default/?path=/story/paragraphs-timeline--default


src/compositions/bcl-timeline/timeline.html.twig
```diff
@@ -0,0 +1,94 @@
...
+{% if _hide_from > 0 %}
+  {% if _hide_to > 0 %}
+    {% set _bottom_amount = _display_amount - _hide_to %}
+  {% else %}
+    {% set _bottom_amount = _hide_to|abs %}
+  {% endif %}
+  {% set _countbase = items|length - _bottom_amount %}
+  {% set _display_amount = _hide_from %}
+  {% set _hide_amount = _countbase - _display_amount %}
+  {% set _toggle_collapsed = _toggle_collapsed|replace({'%d': _hide_amount}) %}
```
planctus, line 47: interesting approach..but it also means that the user will have to pass this "token" for this to work, let's go with it.. :)

src/compositions/bcl-timeline/timeline.html.twig
```diff
@@ -0,0 +1,94 @@
...
+      {# Display the show more button if required. #}
+      {% if (_hide_amount > 0 and loop.revindex == _bottom_amount + 1) or (_hide_amount > 0 and _hide_to == 0 and loop.revindex == 1) %}
+        {% set _toggler = true %}
+        <li class="bcl-timeline__item--toggle">
+          {% include '@oe-bcl/bcl-button/button.html.twig' with {
+            label: "<span class='label-collapsed'>#{_toggle_collapsed}</span><span class='label-expanded'>#{_toggle_expanded}</span>",
+            type: "button",
+            variant: "light",
+            attributes: create_attribute()
+              .setAttribute("aria-expanded", "false")
```
planctus, line 79: i'm trying to get rid of this ambiguity we have in using single and double quotes, normally we use double quotes only when we are writing html, like an attribute, and single quotes when we are dealing with twig code

src/compositions/bcl-timeline/timeline.html.twig
```diff
@@ -0,0 +1,94 @@
...
+          <h6>{{ _item.title }}</h6>
+          {% endif %}
+          <div>{{ _item.content|default('') }}</div>
+      </li>
+      {# Display the show more button if required. #}
+      {% if (_hide_amount > 0 and loop.revindex == _bottom_amount + 1) or (_hide_amount > 0 and _hide_to == 0 and loop.revindex == 1) %}
+        {% set _toggler = true %}
+        <li class="bcl-timeline__item--toggle">
+          {% include '@oe-bcl/bcl-button/button.html.twig' with {
+            label: "<span class='label-collapsed'>#{_toggle_collapsed}</span><span class='label-expanded'>#{_toggle_expanded}</span>",
```
planctus, line 75: mmhh...this would not work. The thing here is that currently in storybook you have this code rendered, but in drupal this string would actually be escaped, it might be better in this case to just write html, not including the button...

src/compositions/bcl-timeline/timeline.story.js
```diff
@@ -0,0 +1,27 @@
...
+    data.icon_path = defaultSprite;
+  }
+
+  return Object.assign(data, args);
+};
+
+export default {
+  title: "Paragraphs/Timeline",
+  parameters: {
+    controls: { disable: true },
```
planctus, line 16: wouldn't it be good to show some controls here, mainly the "hide from" "hide to" to show that this is configurable..?

src/compositions/bcl-timeline/timeline.html.twig
```diff
@@ -0,0 +1,94 @@
...
+  - "toggle_expanded": (string) (default: '')
+  - "hide" (object): (default: undefined)
+    - "from": (integer) (default: items.length) Item index after which to start hiding timeline items
+    - "to": (integer) (default: items.length) Item index after which to resume displaying timeline items
+  - "items" (array) (default: []):
+    - "id": (string) (default: '')
+    - "label": (string) (default: '')
+    - "title": (string) (default: '')
+    - "content": (block) (default: '')
+  - "icon_path" (string) (default: ''): file containing the svg icons
```
planctus, line 16: we have it everywhere as `icons_path`

src/compositions/bcl-timeline/timeline.story.js
```diff
@@ -0,0 +1,27 @@
...
+      {
+        name: "Mockup",
+        type: "figma",
+        url: "https://www.figma.com/file/7aJedLkk8hiDoD3RcxTnQi/BCL-Starter-kit?node-id=6017%3A70268",
+      },
+    ],
+  },
+};
+
+export const Default = (args) => timeline(applyArgs(dataDefault, args));
```
planctus, line 27: why the linter is not complaining..? :)
If you don't use controls, you don't have args, basically, i'm suggesting you to add a few controls (if it's not too complex) and in that case you would need the `applyArgs()` but at the moment you could just override the icons_path anywhere, no need for this function and no need to define args.

planctus (changes requested):
I know that designers already looked at this, so my review is purely on code.. :)

planctus (approved):

PR #122, opened 2021-11-03
feat(landing): Added landing page compositions - FRONT-3329

tibi2303:


src/compositions/bcl-listing/listing.html.twig
```diff
@@ -3,14 +3,18 @@
 {# Parameters:
   - variant: (string) (default: "horizontal") options:
   [ default-1-col, highlight-1-col, default-2-col, default-3-col, highlight-2-col, highlight-3-col ]
-  - title: (string) (default: {})
+  - title: (string) (default: "")
+  - title_tag: (string) (default: "h4")
+  - remove_equal_height: (boolean) (default: false)
```
tibi2303, line 8: I was not sure it was intended to not have equal_height, still i think it is a nice to have here.

src/compositions/bcl-banner/banner.html.twig
```diff
@@ -17,6 +18,18 @@
 {% set _shade = shade|default(false) %}
 {% set _hero = hero|default(false) %}
 {% set _full_width = full_width|default(false) %}
+{% set _content_classes = content_classes|default("") %}
+
+
+{% set _content_class = "bcl-banner__content" %}
```
planctus, line 24: i see a lot confusion, still, about single quotes and double quotes in our templates, i would enforce a style where we use double quotes in the templates when we are "writing html", like an attribute that we set in code will have double quotes, but a variable would have single quotes instead.

src/compositions/bcl-header/header.story.js
```diff
@@ -1,13 +1,14 @@
 import { withDesign } from "storybook-addon-designs";
 import headerData from "@openeuropa/bcl-data-header/data";
 import headerDataEC from "@openeuropa/bcl-data-header/data-ec";
+import headerDataUCPKN from "@openeuropa/bcl-data-header/data-ucpkn";
```
planctus, line 4: i don't get this approach...Why would we demo the ucpkn header in the "default" styleguide..?
This should be done as a "story" override, where you create the story in the ucpkn theme, you exclude the default one and in ucpkn we will simply have this demo, not the ec nor the eu.

src/compositions/bcl-landing-page/landing-page.story.js
```diff
@@ -0,0 +1,67 @@
+import demoData from "@openeuropa/bcl-landing-page/data.js";
+import demoDataUCPKN from "@openeuropa/bcl-landing-page/dataUCPKN.js";
```
planctus, line 2: same here, why would we show the ucpkn landing page in the default theme, and why would we demo the default landing page in the ucpkn styleguide? Also this should be two sories, one for the default and one for ucpkn.

src/compositions/bcl-landing-page/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@openeuropa/bcl-landing-page",
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "0.12.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "description": "OE - BCL landing page",
+  "dependencies": {
```
planctus, line 10: we miss a few dependencies here...

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -3,14 +3,18 @@
 {# Parameters:
   - variant: (string) (default: "horizontal") options:
   [ default-1-col, highlight-1-col, default-2-col, default-3-col, highlight-2-col, highlight-3-col ]
-  - title: (string) (default: {})
+  - title: (string) (default: "")
+  - title_tag: (string) (default: "h4")
```
planctus, line 7: Do we really need to make the listing title tag parametric..?

src/compositions/bcl-listing/listing.html.twig
planctus: you're sure..? How nice? :) 
That class is already "filtered" by the variant of the listing, and it applies only for the highlight variant, where i feel you want to have all the cards aligned so i don't see the need for an addtiional parameter for this.

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -115,9 +119,11 @@
       {% set _card_classes  = [
         'listing-item--highlight',
         'border-0', 'bg-light',
-        'h-100',
         'rounded-2'
       ] %}
+      {% if not _remove_equal_height %}
```
planctus, line 124: as said, i don't see the need for an opt-out and in case i would make it more an opt-in thingy, but i feel that you would always want the "highlight" variant to have all the cards aligned, so...

src/compositions/bcl-landing-page/dataUCPKN.js
```diff
@@ -0,0 +1,187 @@
...
+          },
+          image: {
+            path: "https://picsum.photos/seed/1002/600/400",
+            alt: "alt img",
+            position: "top",
+          },
+        },
+      ],
+    },
+    second_featured_section: {
```
planctus, line 110: well, it's not so important, probably, but couldn't we name these parameters in a more concise and possibly meaningful way..? We presume this "template" will not be used as such, but still it would be better not to have something like` data.second_featured_section.featured_media` in a template.

src/data/header/data-ucpkn.js
```diff
@@ -0,0 +1,25 @@
+const drupalAttribute = require("drupal-attribute");
+
+module.exports = {
+  site_name: "Union Civil Protection Knowledge Network",
+  light: true,
+  project_logo: {
+    path: "ucpkn_logo.svg",
```
planctus, line 7: mmh..not this way. If you are adding the ucpkn logo you have to make it available in the ucpkn theme, first, then you can import it's path form the theme package in the story.

planctus (changes requested):
We miss an .npmignore file here, as we miss it in other compositions, i will create a ticket for the others, here you can do it beforehand ;)

src/compositions/bcl-landing-page/landing-page.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe("OE - Landing page", () => {
+  test("renders correctly", () => {
+    expect.assertions(1);
+
+    return expect(render(demoData)).resolves.toMatchSnapshot();
+  });
+});
+
+describe("OE - Landing page UCPKN", () => {
```
planctus, line 16: am i wrong or you're running both tests with the same data..? :)

src/themes/ucpkn/src/stories/landing-page.story.js
```diff
@@ -0,0 +1,50 @@
...
+    });
+  });
+  data.data.inpage_banner.link.icon.path = defaultSprite;
+  data.data.listing_section.link.icon.path = defaultSprite;
+  data.data.activity_listing.link.icon.path = defaultSprite;
+
+  return data;
+};
+
+export const Platform = () => landingPage(correctPaths(demoData));
```
planctus, line 50: can we show a different label to the user, even default, eh.. :)

src/compositions/bcl-header/header.story.js
```diff
@@ -5,9 +5,9 @@ import header from "@openeuropa/bcl-header/header.html.twig";
 import defaultSprite from "@openeuropa/bcl-bootstrap/bootstrap-icons.svg";
```
planctus, line 5: did you forget to commit a file..? Where is the ucpkn story for the header, now we see the default header instead...

planctus (changes requested):

src/compositions/bcl-header/header.test.js
```diff
@@ -2,6 +2,7 @@ import { renderTwigFileAsNode } from "@openeuropa/bcl-test-utils";
 
 import demoData from "@openeuropa/bcl-data-header/data";
 import demoDataEC from "@openeuropa/bcl-data-header/data-ec";
+import headerDataUCPKN from "@openeuropa/bcl-data-header/data-ucpkn";
```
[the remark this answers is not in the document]
planctus: Yuu forgot about this comment..? You are loading the data for ucpkn but there is not test for it, there should be, instead.

src/compositions/bcl-landing-page/landing-page.test.js
planctus: why now we are not testing the ucpkn version anymore..? :)
The problem was that you were using the wrong data, not the test..

src/themes/ucpkn/src/stories/landing-page.story.js
planctus: Forgot..?

src/compositions/bcl-landing-page/landing-page.test.js
tibi2303: there is only one data for landing page.

src/themes/ucpkn/src/stories/landing-page.story.js
tibi2303: done.

src/compositions/bcl-header/header.test.js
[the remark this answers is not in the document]
tibi2303: done. added the test.

src/compositions/bcl-landing-page/landing-page.test.js
tibi2303: the only thing changed is the text inside the banner.

src/compositions/bcl-landing-page/landing-page.test.js
```diff
@@ -0,0 +1,14 @@
+import { renderTwigFileAsNode } from "@openeuropa/bcl-test-utils";
+
+import demoData from "@openeuropa/bcl-landing-page/data";
+
+const template = "@oe-bcl/bcl-landing-page/landing-page.html.twig";
+const render = (params) => renderTwigFileAsNode(template, params);
+
+describe("OE - Landing page", () => {
```
planctus, line 8: why aren't we testing the ucpkn landing page here..?

planctus (changes requested):
We added the npmignore to all the compositions but here we are still missing it.. :)

planctus (approved):

planctus (approved):

PR #123, opened 2021-11-04
feat(links): Added Underline hover class on link - FRONT-3334

tibi2303:


planctus (changes requested):
Ok, that's it but:

1. i would rename the parameter, though, to `standalone`. 
2. Also the description of the relative control should be improved, what happens is not just that there is an underline on hover, but also and mainly that there is none by default. Say something like "standalone link", although it doesn't mean much :)
3. You created recently an utility css, isn't this the place for this selector you added?
4. i believe that we should also apply this already where we know it should be, i know about the card, the title with a link will have this "new" style

src/components/bcl-link/link.story.js
```diff
@@ -18,6 +18,7 @@ const getArgs = (data, type) => {
     label: data.label,
     path: data.path,
     variant: data.variant,
+    hover_underline: data.hover_underline || false,
```
planctus, line 21: you need to rename it also here

src/components/bcl-link/link.story.js
```diff
@@ -57,6 +58,15 @@ const getArgTypes = (data, type) => {
         category: "Style",
       },
     },
+    standalone: {
+      type: { name: "boolean" },
+      description: "Standalone link",
+      table: {
+        type: { summary: "boolean" },
+        defaultValue: { summary: "false" },
+        category: "Behaviour",
```
planctus, line 67: Put in the existing category "style"

planctus (changes requested):
Almost there.. ;)

planctus (approved):

PR #125, opened 2021-11-08
feat(listing): Listing on all compositions - FRONT-3357

tibi2303:


theosuja (approved):

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -48,19 +59,21 @@
...
+        {% set _card_classes = _card_classes|merge(['px-3']) %}
+      {% endif %}
       {% set _card_attributes = _card_attributes.addClass(_card_classes) %}
       {% set _card = _card|merge({
         horizontal: true,
         extra_classes_body: 'p-0 py-md-0 py-3',
         horizontal_grid: {
-          left_col_classes: 'col-md-3 col-lg-2 rounded ' ~ _max_width,
-          right_col_classes: 'col-md-9 col-lg-10',
+          left_col_classes: 'col-md-3 col-lg-2 rounded ' ~ _max_width ~ (_alignment ? ' align-self-' ~ _alignment : ''),
```
planctus, line 75: well, ok...could we just improve a bit the syntax, like define this class in the upper part of the template together with the other processing of the user data and then use a short ternary operator `?:` here...

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -48,19 +59,21 @@
...
       {% set _card_classes = [
         'listing-item',
-        'px-3',
-        'mb-2',
+        'mb-' ~ _responsive_vertical_spacer ~ '2',
         'border-bottom',
         'border-md-0',
         'border-0',
       ] %}
+      {% if not _remove_horizontal_spacer %}
```
planctus, line 67: i believe we should investigate the usage of this utility, the reason for it, more than using a parameter to avoid it.
I did not even remember about this class added in the default , when i saw it i was suprised..

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -48,19 +59,21 @@
     {% if _variant == 'default-1-col' %}
       {% set _card_classes = [
         'listing-item',
-        'px-3',
-        'mb-2',
+        'mb-' ~ _responsive_vertical_spacer ~ '2',
```
planctus, line 62: mmh..also this margin-bottom surprised me when i saw it..do we need it, responsive or not..?

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -6,19 +6,30 @@
...
+  ['sm', 'md', 'lg', 'xl']
 #}
 
 {% set _items = items|default([]) %}
 {% set _variant = variant|default('default-1-col') %}
 {% set _title = title|default('') %}
+{% set _alignment = alignment|default('') %}
 {% set _link = link|default({}) %}
+{% set _remove_horizontal_spacer = remove_horizontal_spacer|default(false) %}
+{% set _responsive_vertical_spacer = responsive_vertical_spacer|default('') %}
```
tibi2303, line 22: This is mostly used for the case when the image is hidden on mobile.

planctus (changes requested):
I don't know what kind of changes i'#m requesting, the truth is that it seems to me that the listing needs to be tweaked regarding all the things you altered thanks to an additional parameter, the idea is that the listing should work out of the box, zero configuration (well, the variant..), the alignment is the only one that maybe we need to keep but i believe we could get rid of the others...

src/compositions/bcl-listing/listing.html.twig
planctus: do we have such a use case..? :)

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -6,19 +6,30 @@
   - title: (string) (default: {})
   - items: (array of objects of type card)
   - link: (object of type link)
+  - alignment: (string) (default: '') options:
+  ['start', 'end', 'center', 'baseline', 'stretch']
+  - remove_horizontal_spacer: (boolean) (default: false)
+  - responsive_vertical_spacer: (string) (default: '') options:
```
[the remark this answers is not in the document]
tibi2303: I really tried to find another way, this is used when the image is hidden on mobile, the spacing top is too big. Somehow i need to find out if the image is hidden on mobile.
image

src/compositions/bcl-listing/listing.html.twig
tibi2303: https://preview-125--oelibrary.netlify.app/default/?path=/story/compositions-group-listing--list

src/compositions/bcl-listing/listing.html.twig
tibi2303: here is the figma:
image

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -39,55 +42,62 @@
 {% if _items is not empty %}
   {% for _card in _items %}
     {% set _max_width = '' %}
+    {% set _image = {} %}
     {% if _card.image is not empty %}
       {% set _max_width = 'mw-listing-img' %}
+      {% set _image = {
+        path: _card.image.path,
+        classes: _card.image.classes ? _card.image.classes ~ " mb-3" : "mb-3",
+        alt: _card.image.alt ?: "",
```
planctus, line 51: mmhh...i don't get this approach, we have an object with the same keys, why do we "rebuild it"..? I understand you want to set a class  why don't you do it with a |merge..?

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -20,7 +23,7 @@
   type: 'grid',
   responsiveness: 'md',
   gutter: '2',
-  attributes: create_attribute().addClass(['gy-4-5']),
+  attributes: create_attribute().addClass(['gy-4']),
```
planctus, line 26: maybe we can leave this as it was since no one complained about it.. :)

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -39,55 +42,62 @@
...
         'listing-item',
-        'px-3',
         'border-bottom',
         'border-md-0',
         'border-0',
       ] %}
       {% set _card_attributes = _card_attributes.addClass(_card_classes) %}
       {% set _card = _card|merge({
-        extra_classes_body: 'p-0 py-md-0 py-3',
+        extra_classes_body: 'p-0 pb-md-0 pb-3',
```
planctus, line 91: this is problematic:
image

planctus (commented):
a few inline comments left

src/compositions/bcl-listing/listing.html.twig
tibi2303: We should be good now, i missed to add it to the 2 and 3 variants

src/compositions/bcl-listing/listing.html.twig
tibi2303: with 4-5 has the same spacing top and bot

src/compositions/bcl-listing/listing.html.twig
tibi2303: Done. thanks for the help.

planctus (approved):

planctus (approved):

PR #126, opened 2021-11-08
feat(utilities): Add bg-lighter utility. - FRONT-3363

theosuja:
Feel free to share different points of view on how to fix this.


papegaill (commented):
Not that easy to understand the history here of these light gray colors changes...
to me the light color should be #F0F0F0 as on figma default palette colors and gray-100 #F8F9FA
 
please see my remarks on following PR: https://github.com/openeuropa/bootstrap-component-library/pull/124#discussion_r744841537

src/themes/default/src/scss/_buttons.scss
```diff
@@ -0,0 +1,3 @@
+.btn-light {
```
planctus, line 1: I don't understand, really...We have a color palette, then we have this "semantic" colors, like "light", "dark".
This color you are adding shouldn't be now the value of the $light variable and all the utilities generated out of this variable would use it..?
Why overriding the "light" for badges and buttons specifically, instead? The idea is really bad per se, this is a generated utility and we override the selector, not the parameters that generate it...

theosuja: Again: according to the designs we need sometimes elements with a gray `#f0f0f0` background and some other elements with a gray `#f8f9fa` background. From the dev side, we are handling both with the same approach: adding `bg-light` utility. 
- Either we force designers to merge both grays into only one
- Either we use different utility classes, like `bg-light`, `bg-gray` (both already existing) and ` bg-lightgray` or whatever you prefer
- Either we override only the components that are problematic that's what i pretended to do

src/themes/default/src/scss/_variables.scss
```diff
@@ -1,6 +1,6 @@
 // Colors variables
-$gray-100: #f0f0f0;
-$light: $gray-100;
+$light: #f0f0f0; // Make light btns and badges background darker
```
planctus, line 2: would you remove these comments..? i don't find them useful.

src/themes/default/src/scss/_variables.scss
```diff
@@ -23,13 +23,25 @@ $custom-cyans: (
   "cyan-50": $cyan-50,
 );
 
+// Override theme-color "light" and add "lighter"
 $theme-colors: map-merge(
```
planctus, line 27: you know what..? I told you to add this as a theme color, but there are so many selectors created this way and in the end we only need the background utility, not more than that.
So maybe you can remove this merge with the theme colors map, then you keep the rest of the code under the comment "add color lighter to bg utilities" so that we only have this new class available.

src/themes/default/src/scss/_variables.scss
```diff
@@ -23,13 +23,25 @@ $custom-cyans: (
...
   $theme-colors,
   (
     "light": $light,
+    "lighter": $lighter,
   )
-) !default;
+);
 $theme-colors-rgb: map-loop($theme-colors, to-rgb, "$value");
+
+// Add color "lighter" to bg utilities
```
planctus, line 36: There is an "utility.scss" file now in the default theme, this should go there, you can also check the "date" pr, there i moved everything related to utilities in this file.

planctus (changes requested):
Cool, i haven't checked myself all the use cases of this bg-light we had, but now it makes sense, just we don't need all those classes generated out of the theme colors, let's just add bg-lighter

src/themes/default/src/scss/_variables.scss
theosuja: sure!

src/themes/default/src/scss/_variables.scss
theosuja: I tried to only add it to the background utility map, but it looks like the class `bg-lighter` needs the `light` pair to be inside `theme-colors`... otherwise, the `bg-lighter` is `background-color: --bs-lighter-rgb` which is undefined

planctus:
please take also care of the title of this pull request, it will go in the changelog, here we are adding one utlity class, basically, this should be in the title, and something like feat(utlities)

src/themes/default/src/scss/_variables.scss
planctus: too bad, i haven't tested this...uff

planctus: well, i know this will sound silly...but really we don't need these amount of selectors, let's do it dirty, create this `bg-lighter` selector manually in `_utilities.scss` and that's going to be enough..you can protest, eh.. :)

theosuja: go for the silly approach!
It's not silly, it's realistic. I also thought about doing it that way at some point, but I was not sure it would be accepted

src/themes/default/src/scss/_utilities.scss
```diff
@@ -63,3 +63,16 @@ $utilities: map-merge(
     margin-left: -$container-padding-x;
   }
 }
+
+// Override theme-color "light"
+$theme-colors: map-merge(
```
planctus, line 68: sorry..this should go back to the variables file, this is not about utilities, it's a map of colors that we are altering.

src/compositions/bcl-listing/listing.html.twig
```diff
@@ -127,7 +127,7 @@
       {% set _card_classes  = [
         'listing-item--highlight',
         'border-0',
-        'bg-light',
+        'bg-lighter',
```
planctus, line 130: thanks, you checked again, this was just merged.. ;)

planctus (changes requested):
Cool, that's it, move this code and we merge this.. ;)

src/themes/default/src/scss/_utilities.scss
theosuja: ok, my bad!

planctus (approved):

PR #130, opened 2021-11-11
feat(compositions): Add inpage navigation - FRONT-3138

papegaill:
Preview: 
- https://preview-130--oelibrary.netlify.app/default/?path=/story/compositions-inpage-navigation--default

agreed with designers, only the gray background version is covered here, 
the version with transparent background can be used for other projects. this will then be modified via scss by overriding existing styles.

To clarify:

- column content width to large above XL breakpoint


planctus (commented):
You didn't add the css on joinup on purpose..?

planctus (approved):

PR #132, opened 2021-11-15
feat(file): Added file paragraph. - Front 3139

tibi2303:


src/compositions/bcl-file/file.html.twig
```diff
@@ -0,0 +1,146 @@
+{% spaceless %}
+
+{#
+  Parameters:
+  - icon (icon Object) (default: {})
+  - title (string) (default: '')
+  - file (string) (default: '')
```
planctus, line 7: it's not a string, but maybe it should...in ECL it's like that, the param name is "meta", i don't think we need an object, "language" is a param in the root of the component, let's not change these things...

src/compositions/bcl-file/file.html.twig
```diff
@@ -0,0 +1,146 @@
...
+                      .addClass(["text-underline-hover", "align-self-center", "d-inline-block", "mt-1", "mt-md-0"])
+        }) %}
+        {% include '@oe-bcl/bcl-link/link.html.twig' with _download only %}
+      </div>
+    </div>
+  </div>
+  {% if _other_dropdown is not empty %}
+  <div class="bg-light py-3 px-4 border-top">
+    {% if _other_dropdown.alignment is not empty %}
+    <div class="text-{{ _other_dropdown.alignment }}">
```
planctus, line 84: do we need to know the alignment from the data..? I mean, does it need to be parametric..?

src/compositions/bcl-file/file.html.twig
```diff
@@ -0,0 +1,146 @@
...
+    {% endif %}
+    {% include '@oe-bcl/bcl-link/link.html.twig' with _other_dropdown.dropdown_label|merge({
```
planctus, line 86: You really like verbose data, i asked you to rename `other_dropdown` but then did you need to set the label as `dropdown_label`, couldn't it just be `label`

src/compositions/bcl-file/file.html.twig
```diff
@@ -0,0 +1,146 @@
...
+      path: "#" ~ _other_dropdown.id,
```
planctus, line 87: i think we shouldn't ask for `ids` in the data, if we need an id to be set i believe we can define it ourselves with a `random(max)`

planctus (changes requested):
Not sure i looked at everything, we miss the dependency in bcl-twig-templates, eh.. ;)

src/compositions/bcl-file/file.html.twig
tibi2303: Yes, on figma is either on center or the right.

src/compositions/bcl-file/file.html.twig
tibi2303: alignment

src/compositions/bcl-file/file.html.twig
```diff
@@ -0,0 +1,146 @@
...
+
+{# Internal properties #}
+
+{% set _icon = icon|default({}) %}
+{% set _title = title|default('') %}
+{% set _file = file|default('') %}
+{% set _download = download|default({}) %}
+{% set _other_dropdown = other_dropdown|default({}) %}
+{% set _inline_download = inline_download|default(true) %}
+{% set _card = card|default({}) %}
```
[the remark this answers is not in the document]
tibi2303: I am not sure if understood corectly (to add all data for the card at root level of componenent and also make the markup without using the card component) but why overcomplicate stuff... I made the data easier for the user to add it, the grid and the image classes are set by default now.

src/compositions/bcl-file/file.story.js
```diff
@@ -0,0 +1,68 @@
...
+  data.other_dropdown.items.forEach((item) => {
+    if (item.type != "content") {
+      item.download.icon.path = defaultSprite;
+    }
+  });
+
+  return data;
+};
+
+export const Default = () => file(correctPaths(demoData));
```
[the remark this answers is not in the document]
tibi2303: Done, it's for the download link to go under the title and meta of item, not at the end of block.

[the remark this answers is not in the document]
tibi2303: This variant exists in figma.

src/compositions/bcl-file/file.html.twig
```diff
@@ -0,0 +1,146 @@
...
+  }
+#}
+
+{# Internal properties #}
+
+{% set _icon = icon|default({}) %}
+{% set _title = title|default('') %}
+{% set _file = file|default('') %}
+{% set _download = download|default({}) %}
+{% set _other_dropdown = other_dropdown|default({}) %}
```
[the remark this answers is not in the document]
tibi2303: Done.

src/compositions/bcl-file/file.html.twig
```diff
@@ -0,0 +1,146 @@
...
+{% set _inline_download = inline_download|default(true) %}
```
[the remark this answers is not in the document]
tibi2303: It's for the download link to go under the title and meta of item, not at the end of block. I dont think we need a variant here, only a simple boolean.
This variant exists in figma.

src/compositions/bcl-file/file.html.twig
```diff
@@ -0,0 +1,146 @@
...
+{% set _card = card|default({}) %}
+
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+
+{% if _card is not empty %}
+  {% set attributes = attributes.addClass(["border-start", "border-end", "border-top", "rounded-bottom"]) %}
```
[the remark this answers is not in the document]
tibi2303: Indeed, i changed this... :(

src/compositions/bcl-file/file.html.twig
```diff
@@ -0,0 +1,146 @@
...
+      {% if _inline_download %}
+        {% set _container_classes = _container_classes ~ " d-md-flex" %}
+      {% endif %}
+      <div class="{{ _container_classes }}">
+        <div class="flex-grow-1">
+          {% if _title is not empty %}
+            <p class="fw-bold m-0">{{ _title }}</p>
+          {% endif %}
+          <small class="fw-bold m-0">
+            {{ _file.language }}
```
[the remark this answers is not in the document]
tibi2303: I made it the same as on ecl now.

src/compositions/bcl-file/file.html.twig
tibi2303: Same as on ecl now.

src/compositions/bcl-file/file.html.twig
tibi2303: Done.

src/compositions/bcl-file/file.html.twig
tibi2303: The tests fail if it's hardcoded by default. I also add the possibility for the user to add but still generate it by default.

src/compositions/bcl-file/file.html.twig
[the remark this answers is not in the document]
planctus: i actually believe this is just the way the download link would be in mobile, or when there is no space..i don't see the reason why the user would choose one or the other, doesn't make much sense, could you verifiy with the designers..?

src/compositions/bcl-file/file.html.twig
planctus: mmh...i see it on the right only in mobile, why would the user choose between the two, it cannot be..

src/compositions/bcl-file/file.html.twig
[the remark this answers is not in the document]
planctus: So...the concept is basically this: i'm implementing this template and i have my set of data for it, i don't see the reason why, if i want the "card" variant then i have to change my data in order to have it as a card object, what should happen is instead that in our template, if the variant is card we build a card object ourselves and we pass it to the card, but it should be transparent for the user, he is not implementing the card, but the file download.

planctus:
the npmignore is missing

PR #133, opened 2021-11-16
feat(description-list): Added description list - FRONT-2562

tibi2303:


src/compositions/bcl-description-list/description-list.html.twig
```diff
@@ -0,0 +1,53 @@
+{% spaceless %}
+
+{#
```
planctus, line 3: i'm confused, this is called description list but this is actually a template for a single item thingy with a label and a value...
This template should be used to produce a list, and the description list i guess it refers to the html description list:https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

src/compositions/bcl-description-list/description-list.html.twig
```diff
@@ -0,0 +1,53 @@
...
+{% endif %}
+  <div class="d-flex">
+{% if _icon is not empty %}
+  <div>
+  {% include '@oe-bcl/bcl-icon/icon.html.twig' with _icon|merge({
+    attributes: create_attribute().addClass("me-2")
+  }) only %}
+  </div>
+{% endif %}
+    {{- _label -}}
```
planctus, line 36: wouldn't it be better not to expect any markup from the user data..? couldn't we get just the "text" content..?

planctus (changes requested):
I think there was a misunderstanding here :)

planctus:
You keep forgetting to add the component/compositions you add as a dependency in bcl-twig-templates...

src/compositions/bcl-description-list/description-list.html.twig
tibi2303: Remade it the way on ecl, not sure about the extra markup for horizontal though..

src/compositions/bcl-description-list/description-list.html.twig
tibi2303: changed it

planctus (approved):

PR #135, opened 2021-11-17
feat(news): Adding News screens - FRONT-3395

planctus:
About the "style":

- the paragraph margin has been kept as it is in bootstrap, apparently the bigger margin in figma is due to a limitation in the software.
- The sort by label has been moved "inline" compared to the other compositions we have, as it is in figma


.storybook/preview.js
```diff
@@ -53,3 +53,22 @@ export const parameters = {
     },
   },
 };
+
+const scriptInit = (story) => {
+  const demo = story();
+  return `
+    <script>
+      var badges = document.querySelectorAll(".badge");
```
tibi2303, line 61: Nice, we need to also delete it from  group-landing-listing, search, user-listing. I think we can do it in a different task.

src/compositions/bcl-base-templates/sidebar-search.html.twig
```diff
@@ -0,0 +1,7 @@
+{% extends "@oe-bcl/bcl-offcanvas/offcanvas.html.twig" %}
+
+{% block body %}
```
tibi2303, line 3: We need to remove this sidebar from the other compositions and include it from here. I think we need to redo based on content-type template you create everywhere, separate ticket.

src/compositions/bcl-base-templates/content-type.html.twig
```diff
@@ -0,0 +1,94 @@
...
+  - feedback
+  - share
+  - related
+#}
+
+{% set _with_header = header|default(false) %}
+{% set _with_footer = footer|default(false) %}
+{% set _with_banner = banner|default(false) %}
+{% set _with_sidebar = sidebar|default(false) %}
+{% set _columns = columns|default('3-9') %}
```
tibi2303, line 29: I am not sure, but i think in the future we may have different cols, shouldn't this be provide by the user?

src/compositions/bcl-base-templates/listing-page.html.twig
```diff
@@ -0,0 +1,57 @@
...
+{% block content_top %}
+  {% include '@oe-bcl/bcl-button/button.html.twig' with filter_button only %}
+  <div class="mb-3 clearfix">
+    <div class="float-end d-none d-md-flex align-items-center">
+      {% include '@oe-bcl/bcl-select/select.html.twig' with sort_select|merge({
+        attributes: create_attribute().addClass(['ms-2 mb-2']),
+      }) only %}
+    </div>
+    <h4 class="mb-4">
+      <span class="text-capitalize">
```
tibi2303, line 24: I am not sure about this text-capitalize class, I think this should come directly from the data.

src/compositions/bcl-base-templates/listing-page.html.twig
```diff
@@ -0,0 +1,57 @@
...
+    {% if badges is not empty and badges is iterable %}
+      {% for badge in badges %}
+        {% include '@oe-bcl/bcl-badge/badge.html.twig' with badge only %}
+      {% endfor %}
+    {% endif %}
+  </div>
+  <hr class="d-none d-md-block mb-4"/>
+  <div class="d-md-none">
+    {% include '@oe-bcl/bcl-select/select.html.twig' with sort_select|merge({
+      extra_classes: "mb-4"
```
tibi2303, line 38: Double quotes :D

src/compositions/bcl-base-templates/sidebar-search.html.twig
planctus: yep, sure, that's the idea, but since it seems that this "content-type.html.twig" is actually the base template potentially for all the compositions, i will rename it to like `basic-page.html.twig`

src/compositions/bcl-base-templates/content-type.html.twig
planctus: Which user..? :) I think this is a "utility" template for us, i don't think this is anyhow going to be used by anyone else..

src/compositions/bcl-news/bcl-news.story.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+  if (data.filter_button) {
+    data.filter_button.icon.path = defaultSprite;
+  }
+  if (data.badges) {
+    data.badges.forEach((badge) => {
+      badge.icons_path = defaultSprite;
+    });
+  }
+  if (data.footer) {
```
tibi2303, line 59: This can be tackled in a different ticket, but everywhere we have a footer or a header in the composition, we need to add  this, i am not sure if there is a way to add it on all stories from one source.

src/compositions/bcl-base-templates/listing-page.html.twig
planctus: Why..? we pass content_type and we pass it lowercased, we also use it to generate classes, that class is harmless, even if you have it capitalized from the source..

src/compositions/bcl-news/bcl-news.story.js
planctus: not too hard to do it let's add this to the story-utils. as you said in another ticket

.storybook/preview.js
planctus: yeah, follow-up ticket would be to use this approach for all the existing compositions, as we did for the listing...

src/compositions/bcl-base-templates/content-type.html.twig
tibi2303: Sure, i just hope we won't have multiple variants on this 2-10, 1-11 etc..

src/compositions/bcl-base-templates/listing-page.html.twig
tibi2303: Ok, indeed

PR #142, opened 2021-11-24
feat(paragraphs): Homogenize titles across all Paragraphs. - FRONT-3418

theosuja:
As discussed, the breaking change is the switch in the naming of `title` variable in the `file` paragraph. The former one becomes `item_title` and `title` will now be used as an optional "section title" to keep consistency between all paragraphs.

### DIFF
--- a/src/compositions/bcl-banner/__snapshots__/banner.test.js.snap
+++ b/src/compositions/bcl-banner/__snapshots__/banner.test.js.snap
@@ -40,6 +40,46 @@ exports[`OE - Hero banner default renders correctly 1`] = `
 </jest>
 `;
 
+exports[`OE - Hero banner default renders correctly with title 1`] = `
+<jest>
+  <div
+    class="bg-lighter text-dark text-center bcl-banner  hero"
+  >
+    <div
+      class="container"
+    >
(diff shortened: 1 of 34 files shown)


planctus:
It seems fine, can i ask you to:

- add the title in the test of the paragraphs
- add a control in all the related stories (can be with default empty)
- Ensure that all the instances are exactly the same (bold, margin-bottom)

theosuja:
> It seems fine, can i ask you to:
> 
>     * add the title in the test of the paragraphs
> 
>     * add a control in all the related stories (can be with default empty)
> 
>     * Ensure that all the instances are exactly the same (bold, margin-bottom)

* Tests updated, found a couple of pre-existing bugs and fixed them by the way.

* Controls have been added except for the stories that explicitly have disabled controls. Note that I struggled particularly with the `listing.story.js` so it could require extra attention when reviewing.

* As discussed, this point is not going to be implemented since it would break the visual balance of components with a specific title (such as the banner, links block, and some more...). All paragraphs without an intrinsic title have been provided with the same title as the `listing` paragraph.

tibi2303 (approved):

PR #152, opened 2021-11-30
feat(badges): Added outline "variant" - FRONT-3429

tibi2303:


src/components/bcl-badge/badge.html.twig
```diff
@@ -8,13 +8,15 @@
   - dismissible (boolean) (default: false)
   - assistive_text (string) (default: '')
   - rounded_pill (boolean) (default: false)
+  - outline (boolean) (default: false)
   - icons_path (string) (default: '')
 #}
 
 {% set _rounded_pill = rounded_pill|default(false) %}
```
planctus, line 15: question:
couldn't we introduce the variant parameter..? We now have default, pill and outline, i feel this would be the way to handle this. But this would introduce a breaking change, at least because if the rounded_pill parameter is "used" in the drupal paragraph. We can anticipate this to the developers and see if they can handle this in the next release.

src/themes/default/src/scss/_badge.scss
```diff
@@ -83,3 +83,11 @@ a {
     font-size: 0.75em; // Default size in Bootstrap
   }
 }
+
+@each $color, $value in $theme-colors {
```
planctus, line 87: mmh..this is not enough, if you check the styles the "circle" has the same background color of the border, the x can be white or kind of black...

planctus (changes requested):
I think you forgot some data file, like the content banner ones maybe, please double check this..

papegaill (changes requested):
Thanks 👍  here is a few small remarks for the users profile pages.

**Listing:**
Screenshot 2021-12-02 at 10 27 13
missing little space above the pagination

**Edit:**
Screenshot 2021-12-02 at 10 27 39
should  be light gray

Screenshot 2021-12-02 at 10 27 48
missing space bellow the button

**View:**
Screenshot 2021-12-02 at 10 27 59
should  be light gray

Screenshot 2021-12-02 at 10 28 15
missing space bellow the content

planctus (changes requested):
I still see the normal badges used  in many stories, think i've asked to double check previously, i believe it should be easy, considering that everyhwere i see badges in figma they are pills and with the outline...

planctus (commented):
can we add some margin of the default card story..?
Actually we could also remove the button there, it looks bad there together with badges..

tibi2303:
Done @planctus

planctus (approved):

PR #153, opened 2021-11-30
feat(icons): Update bootstrap icons - FRONT-3437

tibi2303:


planctus:
i don't get it...what did you do..? :)
If you want to update bootstrap-icons, that is an npm package set as a dependency of bcl-bootstrap, but in any case we don't perform "single" updates, i mean that being like any other package it is part of the "normal" updates. The sprite is copied in each theme through the builder and i realised today that we were versioning these files that we copy, so i added the icons in the gitignore file in another pr. I think we can close this pr

tibi2303:
@planctus I did a copy paste from the bootstrap sprite icons into bootstrap/bootstrap-icon.svg, The builder added it into the themes. Sure we can have the sprites from the themes in ignore but how should we proceed with the update of the sprite? That is how i created the sprite in the beginning, just a simple copy from bootstrap icons.

planctus:
The update of the bootstrap icons sprite is done by updating the dependency in bcl-bootstrap as i told you in the previous comment.
You might be confused, please check the `bcl-bootstrap` package and you will notice that the sprite file is copied into the root of the package via the `copy-icons` script.
I thought the previous comment would have clarified this, i think you can close this pull request basically, the update of the icons, as i told you already, is done together with other updates, being nothing else that an npm dependency.

PR #154, opened 2021-11-30
feat(flags): Adding flag example in icon, distributing sprite with the themes - FRONT-3430

planctus:


tibi2303:
Very nice, still I think we can add some tests for it too.

planctus:
what test..? :dagger: it's just "names", "strings" which is user data, the component is already tested..so

tibi2303 (approved):

PR #157, opened 2021-12-02
feat(header): Adding "neutral" variant - FRONT-3431

planctus:


tibi2303:
image
I don't think this solution work out for all cases, on ucpkn for example.

src/data/header/data-ucpkn.js
```diff
@@ -1,6 +1,7 @@
 const drupalAttribute = require("drupal-attribute");
 
 module.exports = {
+  variant: "",
```
tibi2303, line 4: Here should be an empty variant?

tibi2303 (approved):

PR #160, opened 2021-12-06
feat(footer): Added footer neutral - OEL-827

tibi2303:


planctus (changes requested):
You said you would have looked at what has been done for the header...I don't see the reason why header and footer would require an "element" like `ec__` or `eu__ `footer would be as the wrapper class.
I made modifiers in header, i think we can do the same for the footer `bcl-footer--ec`, `bcl-footer--eu` and these classes need to be set by the template, create the variant parameter and set it neutral as default.
In the sass you will keep this "deprecated" class not to introduce breaking changes.

planctus (approved):

PR #161, opened 2021-12-06
fix(inpage-navigation): Changing selector used for styling - OEL-839

tibi2303:


src/compositions/bcl-inpage-navigation/data--simple.js
```diff
@@ -1,7 +1,11 @@
 const drupalAttribute = require("drupal-attribute");
 
 module.exports = {
-  attributes: new drupalAttribute().addClass(["d-none", "d-lg-block"]),
+  attributes: new drupalAttribute().addClass([
+    "d-none",
+    "d-lg-block",
+    "bcl-navigation",
```
planctus, line 7: bcl-inpage-navigaiton, but this has to happen in the template, not by an attribute added via the user data, this is the component class.

src/themes/default/src/scss/_inpage-navigation.scss
```diff
@@ -1,4 +1,4 @@
-#bcl-inpage-navigation {
+.bcl-navigation {
```
planctus, line 1: the selector was correct, just it has to be turned into a class and not an id

planctus (approved):

PR #162, opened 2021-12-06
feat(layout): Adaptive layout approach - OEL-828

planctus:


tibi2303:
image
I think it's my fault here, should be d-md-none on button.

tibi2303 (approved):

PR #181, opened 2021-12-16
feat(dropdown-header): Fix for long dropdown items inside header - OEL-891

tibi2303:


src/themes/default/src/scss/_header.scss
```diff
@@ -144,6 +144,22 @@ $header-link-active-background: #003776 !default;
       justify-content: right;
     }
   }
+
+  .bcl-header__navbar .navbar-collapse {
+    max-width: 100%;
+  }
```
planctus, line 150: can we indent something here..? I mean, it's sass :)

tibi2303: done.

PR #182, opened 2021-12-16
feat(ucpkn-header): Fix for overlapping site name - OEL-889

tibi2303:


planctus:
mmhh..yeah, we fix the problem of the overlap, but it looks to me that the position of the button, meaning its alignment is not correct, it should be centered, i believe. The thing is that the positioning of that button was already "problematic", but i don't know if you can refactor that not to be "absolute".

PR #183, opened 2021-12-20
fix(featured-media): Fix media caption size with different image size - OEL-910

papegaill:


tibi2303:
I am not sure about this approach. If width of the figcaption can now be set by the width of text. ex:
image
image


Is this correct?

tibi2303 (approved):

PR #185, opened 2021-12-20
feat(search): Updates on search page - OEL-914

tibi2303:


src/components/bcl-button/button.html.twig
```diff
@@ -15,6 +15,8 @@
   - icon (icon object) (default: {}): predefined structure for Icon component
   - icon_position (string) (default: after)
     - options ["after", "before"]
+  - icon_spacing_responsive (string) (default: '')
```
planctus, line 18: can we use short strings for these parameters name..? It doesn't have to be fully descriptive of what that is about..

planctus (changes requested):
fix this and then i will be able to take a look at it: https://preview-185--oelibrary.netlify.app/default/?path=/story/pages-search--default

planctus (approved):

PR #187, opened 2021-12-21
feat(accessibility): Select & Multiselect report - OEL-918

tibi2303:
The Incomplete rule is because of the arrow bg image on the right of the select. I think it should not be taken into consideration.

Multiselect
The only violation remained is:
Scrollable region must have keyboard access

But i think it's not necessary. that violation is regarding the select that we hide it using opacity 0 and the rule apply to it even if it's hidden (should have been done with display:none; to pass but this approach was done because of the focus state of multiselect #78 ) and it also checks the list of the multiselect that can not be controlled using the tab (which this is checked in the rule) but it can be controlled using the arrows so i think it is good enough.

### DIFF
--- a/src/components/bcl-select/select.html.twig
+++ b/src/components/bcl-select/select.html.twig
@@ -2,6 +2,7 @@
 
 {#
   Parameters:
+  - "id" (string) (default: '')
   - "label" (string) (default: '')
   - "hidden_label" (boolean) (default: false)
   - "helper_text" (string) (default: '')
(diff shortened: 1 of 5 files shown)


src/components/bcl-select/select.html.twig
```diff
@@ -55,6 +57,10 @@
   {% set attributes = attributes.setAttribute('aria-label', _aria_label) %}
 {% endif %}
 
+{% if _id is not empty %}
```
planctus, line 60: do we need this..?
why not simply passing it as an attribute in our stories..?

PR #191, opened 2021-12-22
feat(accessibility): Pagination icons report - OEL-922

tibi2303:


src/components/bcl-pagination/pagination.html.twig
```diff
@@ -75,6 +75,9 @@
         tabindex="-1"
         aria-disabled="true"
         {% endif %}
+        {%- if _enable_icon -%}
+        aria-label="previous"
```
planctus, line 79: mmhh..it's an aria attribute, ok, but still we cannot hardcode strings in a template, this should be an additional parameter, most likely..

planctus (changes requested):
See the comment inline

PR #199, opened 2022-01-04
feat(accessibility): Group report - OEL-929

tibi2303:


src/compositions/bcl-group/dataListContent.js
```diff
@@ -61,33 +63,30 @@ module.exports = {
...
-            legend: "Creation date (to)",
-            fieldset_classes: "mb-3",
-            legend_classes: "col-form-label",
-          },
           {
             input_type: "date",
+            label: "Creation date (to)",
+            wrapper_classes: "mb-3",
             attributes: new drupalAttribute().addClass("mb-2"),
+            id: "creation-date=to",
```
planctus, line 89: i guess this was supposed to be `creation-date-to`

tibi2303: yeah, my mistake, done.

PR #208, opened 2022-01-07
feat(logos): Added scripts - OEL-953

tibi2303:


planctus:
I don't think you did this on purpose, or maybe you did, but in EU what we are doing is to copy all the svg in the different folders, the files have the same name and the result is that the only files we have are the "standard" version.
Now, i'm not sure it's really a good idea to ship all these files in the theme packages, as you know, but if we have to it might be better to copy everything.

.gitignore
```diff
@@ -9,10 +9,13 @@ bootstrap/js
 bootstrap/scss
 src/themes/*/css
 src/themes/*/js
+src/themes/*/logos
 src/themes/**/bcl-stories
 src/themes/ecl/storybook/.storybook/*.js
 src/themes/joinup/storybook/.storybook/*.js
 src/themes/ucpkn/storybook/.storybook/*.js
+src/themes/ucpkn/logo/ec
```
planctus, line 17: couldn't you just do this on one line with
`src/themes/*/logo*`

src/themes/joinup/bcl-builder.config.js
```diff
@@ -120,6 +120,21 @@ module.exports = {
       to: path.resolve(outputFolder, "icons"),
       options: { up: true },
     },
+    {
```
planctus, line 123: it's probably my fault that i told you that it would not have possible to do everything in one operation for EC, but why not..? :)
`from: [`${nodeModules}/@ecl/resources-ec-logo/**/*.svg`],`
will get both the files in the root and the ones in the nested folders, basically.. ;)

planctus (changes requested):

src/themes/joinup/bcl-builder.config.js
tibi2303: Right now i splitted them into positive, negative. The root ones are positive and the one in the folder are entirely copied with the folder (folder being named negative) inside theme. If we do this we will have the same structure the package has (which in my opinion is not the best way). If you want i can go like this.

.gitignore
```diff
@@ -9,10 +9,12 @@ bootstrap/js
 bootstrap/scss
 src/themes/*/css
 src/themes/*/js
+src/themes/*/logos
```
planctus, line 12: the one you added now is enough..this is not needed anymore

src/themes/ucpkn/package.json
```diff
@@ -9,11 +9,13 @@
     "build:styles": "cross-env bcl-builder styles",
     "build:scripts": "cross-env bcl-builder scripts",
     "build:copy": "cross-env bcl-builder copy",
-    "clean": "rimraf css js templates",
+    "clean": "rimraf css js templates logos",
```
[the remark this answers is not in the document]
tibi2303: yeah, actually should be logo/ec logo/eu as i need the folder for that ucpkn logo i added manually

planctus (changes requested):
small things..

src/themes/ucpkn/package.json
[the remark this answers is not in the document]
tibi2303: done.

.gitignore
tibi2303: done.

src/themes/ucpkn/package.json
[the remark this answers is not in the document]
planctus: but why do you want to keep the empty logo folder, i don't get it...? :D

planctus (approved):

PR #212, opened 2022-01-10
feat(date-block): Revamp composition - OEL-988

papegaill:


src/compositions/bcl-date-block/date-block.html.twig
```diff
@@ -38,9 +38,8 @@
 {% endif %}
 
 <time {{ attributes }}>
-  <span>{{ _day }}</span>
-  <span class="pb-3 text-uppercase">{{ _month }}</span>
-  <span class="bg-light w-100 text-center py-2 text-dark rounded-bottom mb-n1">{{ _year }}</span>
+  <span class="py-4 bg-white border border-bottom-0 w-100 fs-5 fw-medium text-dark text-center text-uppercase">{{ _day }} {{ _month }}</span>
+  <span class="bg-light border w-100 text-center fs-5 text-dark rounded-bottom mb-n1">{{ _year }}</span>
```
tibi2303, line 42: Not the correct background, should be lighter

src/themes/default/src/scss/_utilities.scss
```diff
@@ -23,7 +23,7 @@ $utilities: map-merge(
             map-merge(
               map-get(map-get($utilities, "max-width"), "values"),
               (
-                date: 7.5rem,
+                date: 8.5rem,
```
tibi2303, line 26: I properly measure it on figma (because of the license i think) but it seems bigger in the implementation.

tibi2303 (approved):

PR #218, opened 2022-01-12
feat(color-overwrite): Fix for color variables overwrite - OEL-893

tibi2303:


src/themes/joinup/src/scss/oe-bcl-joinup.scss
```diff
@@ -1,9 +1,12 @@
+// Custom theme colors
+@import "@openeuropa/bcl-theme-joinup/src/scss/base/colors";
+
 // Configuration
 @import "@openeuropa/bcl-bootstrap/scss/functions";
 @import "@openeuropa/bcl-bootstrap/scss/variables";
 
 // Override Bootstrap variables
-@import "@openeuropa/bcl-theme-default/src/scss/variables";
+@import "@openeuropa/bcl-theme-default/src/scss/base/variables";
```
planctus, line 9: my question is: do we need this base folder..? Isn't it just an overhead..?

tibi2303: i am thinking of having the same structure over all themes even if it's default.

planctus: i understand, fine, but i meant that we probably don't need a "base" subfolder in any of the themes.. :)

planctus (approved):

PR #225, opened 2022-01-14
fix(header): Review button display and icon alignement - OEL-1023

papegaill:


src/themes/default/src/scss/_button.scss
```diff
@@ -0,0 +1,5 @@
+// Button
+.btn {
```
tibi2303, line 2: It seems to work in most places, hard to check every button.

tibi2303 (approved):

src/themes/default/src/scss/_button.scss
tibi2303: Checked possible solutions on button component. seems to work ok.

PR #232, opened 2022-01-20
feat(contact-form): Adding examples - OEL-1055

planctus:


src/components/bcl-alert/__snapshots__/alert.test.js.snap
```diff
@@ -14,7 +14,7 @@ exports[`OE - Alert danger renders correctly 1`] = `
       />
     </svg>
     <div
-      class="alert-content"
+      class="alert-content flex-grow-1"
```
papegaill, line 17: `.flex-grow-1` seems useless?

src/components/bcl-form/form.html.twig
```diff
@@ -98,6 +117,13 @@
   {% endif %}
 {% endfor %}
 {% endif %}
+{% if _antispam is not empty %}
+  <div class="mt-4 mb-2">
+    <div class=" bg-light">
```
papegaill, line 122: space to be removed before the class `.bg-light`

src/compositions/bcl-contact-form/contact-form.story.js
```diff
@@ -0,0 +1,85 @@
...
+        name: "Mockup - Contact form",
+        type: "figma",
+        url: "https://www.figma.com/file/NQlGvTiTXZYN8TwY2Ur5EI/BCL-Features?node-id=5191%3A180830",
+      },
+    ],
+  },
+};
+
+export const Contact = () => contactForm(correctPaths(dataDefault));
+console.log({ ...baseData, dataDefault });
```
papegaill, line 75: `console.log` to be removed

papegaill (changes requested):

src/compositions/bcl-base-templates/content-type.html.twig
```diff
@@ -70,6 +71,16 @@
 {% endif %}
 
 <main {{ attributes }}>
+{% if messages is defined and messages is not empty %}
+  {% block messages_area %}
+  <div class="container mt-lg-5 mt-2">
```
papegaill, line 76: maybe margin classes have to be set on the alert itself, when removing the alert message, the container stays empty with these spacing classes.

src/components/bcl-alert/__snapshots__/alert.test.js.snap
planctus: This is without...it's flex, so the alert-content doesn't use all the available room if not needed.
image

src/compositions/bcl-base-templates/content-type.html.twig
planctus: yep, you're right..

papegaill (approved):

PR #234, opened 2022-01-21
feat(event): Revamp to v1.1 - OEL-1056

papegaill:


src/compositions/bcl-description-list/__snapshots__/description-list.test.js.snap
```diff
@@ -61,12 +61,8 @@ exports[`OE - Description List horizontal renders correctly 1`] = `
           Only title label
     </dt>
     <div>
-      <dd>
-        Description text goes here.
-      </dd>
-      <dd>
-        Description text goes here.
-      </dd>
+      <dd />
```
planctus, line 64: are these changed expected alexis..?

papegaill: Thks, It fixed now. I had checked in @tibi2303 's Project PR to follow the same approach for the flag icons in the description-list.. it will probably need to be updated as well.

tibi2303: we will conflict i think but it's an easy fix.

papegaill: Sure but you have same issue in your PR, some description of the description-list are now empty ;)

tibi2303: Yeah, thanks, i noticed now. It's because of me, i added the possibility to include the icons/buttons/links inside that label but forgot to update the data for horizontal. 
inside description-list/dataHorizontal.js
From this:
`
      definition: [
        "Description text goes here.",
        "Description text goes here.",
      ],`
should be this
`
      definition: [
        {
          label: "<p class='mb-2'>Description text goes here.</p>",
        },
        {
          label: "<p class='mb-2'>Description text goes here.</p>",
        },
      ],`

papegaill: ok then should be updated in vertical data as well no?

tibi2303: yes indeed.

src/compositions/bcl-event/event.story.js
```diff
@@ -1,9 +1,13 @@
 import { withDesign } from "storybook-addon-designs";
-import { initBadges, correctPaths } from "@openeuropa/bcl-story-utils";
-import sidebar from "@openeuropa/bcl-event/data/event-sidebar.js";
+import {
+  initScrollspy,
```
planctus, line 3: we need to update the link to the figma file.. ;)

src/compositions/bcl-event/event.test.js
```diff
@@ -26,17 +26,7 @@ const baseData = {
...
-  },
-  banner: {
-    title: {
-      content:
-        "Translation, Interpreting, Culture 2: Human Factor in translation Technologies",
-      tag: "h2",
-    },
-  },
+  ...dataPage,
   files: files,
```
planctus, line 30: it's not something you touched here, but i wonder the reason why now we have red pdf icons.. :) Can we fix this, it's the text-secondary class that is added directly in the file template, it's clear that we cannot use a color from the theme colors  for something like this, should be i guess the same color for all the sub-themes, unless differently specified..

planctus:
would you add:
```
badgesConfig: {
      stable: {
        title: "v1.1",
      },
    },
```
in the parameters so we show the v1.1 badge for this feature..? To me the implementation is good, i've found issues in the design while checking it, and indeed i don't see needed changes on our side (the listing looks really different, though), just maybe we could take the chance to fix these file icons, or we do it in a different ticket.

planctus (commented):
Two small updates and i let you evaluate whether quickly fixing the file icons in this pr or we can create a ticket for that. ;)
Thanks!

src/compositions/bcl-event/data/data-page.js
```diff
@@ -0,0 +1,87 @@
...
+      },
+    ],
+  },
+  details_list: {
+    attributes: new drupalAttribute().addClass(["mt-3", "mb-4"]),
+    variant: "horizontal",
+    items: [
+      {
+        term: {
+          label: '<p class="mb-0 d-inline-block">Date</p>',
```
planctus, line 56: since i see these classes being added consistently to achieve a certain layout, couldn't this happen in the template basing on a "parameter" instead of needing custom markup to be injected from the data..?
Another possibility could also be to make this a template on its own, as a "drupal block" template, a composition in our library, where you would add these classes in the template..

planctus (commented):
did the designers already approve this, @papegaill..? I would just "improve" the new "block" making it easier to "implement" it, if you feel to ;)

planctus:
to make the badge for the versioning "visible" we also need to declare it as a parameter, i only gave you the code to alter its configuration, so in the parameters object please add also:
`badges: ['stable],`

planctus (approved):

PR #240, opened 2022-01-25
feat(accordion): Added accordion title_tag - OEL-1071

tibi2303:


src/data/accordion/data.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   id: "1",
   items: [
     {
+      title_tag: "h3",
```
planctus, line 5: let's keep the demo as it is, if you want to "test" this parameter you can do it in the jest file

tibi2303: done.

planctus (approved):

PR #245, opened 2022-02-01
feat(multilingual): Added feature - OEL-1084

tibi2303:


src/components/bcl-twig-templates/package.json
```diff
@@ -37,6 +37,7 @@
     "@openeuropa/bcl-icon": "^0.19.0",
     "@openeuropa/bcl-inpage-navigation": "^0.19.0",
     "@openeuropa/bcl-landing-page": "^0.19.0",
+    "@openeuropa/bcl-language": "^0.19.0",
```
planctus, line 40: i don't understand the naming. i would name the "main" package as the feature is: bcl-multilingual and its story should be in the Features group for consistency. This demo for me should show a standard page with the language list modal opened, like if the user clicked already on the language switcher button.

I would then separate the "story" from the twig package, i mean that bcl-multilingual could basically only contain the story and the needed data, not even being a public package, and i would have a **language-list** composition with templates named:
**language-list.html.twig**
**language-list-overlay.html.twig**

In ECL there is a template for the item that is called from the main template, but i don't think we need to reproduce that, but i would expected this template to be based a bit more (when possible) on the data structure of the ECL component including the parameters name.
Did you take into account the ECL version of this "feature"..? Because surely this exists already in drupal, using that template, so the more we limit the changes the best is, i think..

src/compositions/bcl-language-list/language-list-modal.html.twig
```diff
@@ -0,0 +1,58 @@
...
+      </div>
+    </div>
+    {% endif %}
+  </div>
+  {% endif %}
+{% endblock %}
+
+{% block body %}
+  {% include '@oe-bcl/bcl-language-list/language-list.html.twig' with {
+    attributes: create_attribute()
```
planctus, line 56: mmh...why this? To reset them..? But these attributes once sent back to the language-list template would not be used, right? so it doesn't look like we need this..

src/compositions/bcl-language-list/language-list-modal.html.twig
```diff
@@ -0,0 +1,58 @@
...
+      <div class="py-3-5 px-3 px-sm-0 px-md-3">
+    {% endif %}
+    <div class="d-flex justify-content-between align-items-center">
+  {% endif %}
+  <div class="d-flex align-items-center">
+    {% set _title_classes = 'mb-0' %}
+    {% if _icon_path is not empty %}
+      {% set _title_classes = _title_classes ~ ' ms-2' %}
+      {% include '@oe-bcl/bcl-icon/icon.html.twig' with {
+        name: "globe",
```
planctus, line 33: single quotes

src/compositions/bcl-language-list/language-list.html.twig
```diff
@@ -0,0 +1,114 @@
...
+      {% endif %}
+      {% if not (_variant == 'neutral' and _inside_container) %}
+      </div>
+      {% endif %}
+    </div>
+    {% endif %}
+  </div>
+{% endif %}
+
+
```
planctus, line 113: one empty line is enough.. :)

src/compositions/bcl-language-list/language-list.story.js
```diff
@@ -0,0 +1,32 @@
...
+import { correctPaths } from "@openeuropa/bcl-story-utils";
+import demoData from "@openeuropa/bcl-language-list/data/data--neutral.js";
+import demoDataEU from "@openeuropa/bcl-language-list/data/data--eu.js";
+import demoDataEC from "@openeuropa/bcl-language-list/data/data--ec.js";
+import language from "@openeuropa/bcl-language-list/language-list.html.twig";
+
+export default {
+  title: "Compositions/Language List",
+  decorators: [withDesign],
+  parameters: {
```
planctus, line 11: we don't have controls, can we disable the panel then..?

src/compositions/bcl-language-list/language-list.test.js
```diff
@@ -0,0 +1,28 @@
...
+    expect.assertions(1);
+
+    return expect(render(demoDataEU)).resolves.toMatchSnapshot();
+  });
+
+  test("EC renders correctly", () => {
+    expect.assertions(1);
+
+    return expect(render(demoDataEC)).resolves.toMatchSnapshot();
+  });
```
planctus, line 27: shouldn't we add a test also for the modal.?

planctus (changes requested):
Small things, thanks a lot for this work, this is a big pull request, it seems fine to me now.. ;)

src/compositions/bcl-language-list/language-list-modal.html.twig
tibi2303: image
They are being used after the base template it's called and also i need to reset them in order to clear the classes i put on the modal here:
image

src/compositions/bcl-multilingual/.npmignore
```diff
@@ -0,0 +1 @@
+*.js
```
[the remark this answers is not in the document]
tibi2303: done.

src/compositions/bcl-language-list/language-list.html.twig
```diff
@@ -0,0 +1,114 @@
...
+  - overlay (boolean) (default: false)
+  - inside_container (boolean) (default: false)
+  - variant (string) (default: 'neutral')
+  - title (string) (default: '')
+  - logo (object) (default: {}) format: {
+      path (string),
+      alt (string),
+      class (string),
+    }
+  - icon_path (string) (default: '')
```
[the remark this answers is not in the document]
tibi2303: done.
I will create a ticket regarding this, i think we should change all of them into 'icon_path'

src/compositions/bcl-language-list/language-list.html.twig
```diff
@@ -0,0 +1,114 @@
...
+    - "active (boolean) (default: false)
+  - non_eu_items (array) (default: [])
+    - "lang" (string) (default: '')
+    - "label" (string) (default: '')
+    - "path" (string) (default: '')
+    - "active (boolean) (default: false)
+  - close_label (string) (default: '')
+#}
+
+{% set _id = id|default('') %}
```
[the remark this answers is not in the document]
tibi2303: yes, thank you. I deleted it.

src/compositions/bcl-language-list/language-list-modal.html.twig
tibi2303: done.

src/compositions/bcl-language-list/language-list.html.twig
tibi2303: done.

src/compositions/bcl-language-list/language-list.story.js
tibi2303: done. Also disabled them in Multilingual too

src/compositions/bcl-language-list/language-list.test.js
tibi2303: They are already tested in header but i added in here too for each variant.

planctus (approved):

PR #246, opened 2022-02-02
feat(blocks): Added accordion, alert, badge, breadcrumb block - OEL-1108

tibi2303:


src/components/bcl-accordion/accordion.html.twig
```diff
@@ -56,11 +59,13 @@
                                 .setAttribute('aria-controls', 'collapse-' ~ _id ~ '-' ~ loop.index)
                                 .setAttribute('aria-expanded', open_item ? 'true' : 'false')
       %}
-      {% include '@oe-bcl/bcl-button/button.html.twig' with {
-        label: _item.title,
-        clean_class: true,
-        attributes: button_attributes
-      } only %}
+      {% block accordion_button %}
+        {% include '@oe-bcl/bcl-button/button.html.twig' with {
```
planctus, line 63: mmh..didn't we say that all the processing should still be available "out of the block"..? So that whatever is the data processing we perform it will still be used but then a different template can be included...

planctus (changes requested):
i know i told you you could have made changes in "bulk", but maybe it's better of we go one by one, otherwise you see that now we have to apply multiple times the same thing to different template. Let's reach a clear understanding of what and how we are doing it (also let the backend devs validate this) and only after this we can go faster grouping things together..

PR #255, opened 2022-02-08
feat(project): Add circular progress bar for project status - OEL-1137

papegaill:


src/compositions/bcl-project-status/project-status.html.twig
```diff
@@ -99,11 +103,26 @@
     </div>
   </div>
 </div>
-
-{% if _charts is not empty %}
+{% if _chart_progress is not empty and _chart_details_list is not empty %}
 <div class="border-bottom pb-4 mb-4">
-  {%- block charts _charts -%}
+  {% block charts %}
+  <div class="row">
```
tibi2303, line 109: As we talked i think it would be better to have this behaviour in the progress component. I know it can be a bit tricky, but i find this to be only a variant to that component.

src/themes/default/src/scss/_circular-progress-bar.scss
```diff
@@ -0,0 +1,147 @@
...
+    -webkit-transform-origin: center right;
+    transform-origin: center right;
+    //animation: loading-1 1.8s linear forwards;
+  }
+}
+
+.circular-progress-value {
+  display: flex;
+  border-radius: 50%;
+  font-size: 36px;
```
tibi2303, line 77: This can be added as variables, or even use the bs variables.

tibi2303 (changes requested):
You can find some points I added here.

src/themes/default/src/scss/_circular-progress-bar.scss
tibi2303: I am reffering to all hardcoded values.

src/compositions/bcl-project-status/project-status.html.twig
planctus: ideally this could be indeed a "variant" of the progress bootstrap component, but i don't think we need it as such, extending a bs component. We just have this use case here and it's probably better to keep it like this.
What i believe is that this project-status deserves now a story to be added, where we can also "expose" the relevant controls. ;)

src/themes/default/src/scss/_circular-progress-bar.scss
planctus: so let's remove it.. :)

planctus: so let's remove it.. :)

src/themes/default/src/scss/_circular-progress-bar.scss
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+//animation
+@for $i from 1 through ($howManySteps / 2) {
+  $degrees: (180 / ($howManySteps / 2));
+  $degrees: ($degrees * $i);
+  @keyframes loading-#{$i} {
+    0% {
+      -webkit-transform: rotate(0deg);
```
planctus, line 139: also this should not be needed

src/themes/default/src/scss/oe-bcl-default.scss
```diff
@@ -54,6 +54,7 @@
 // Custom styles
 @import "@openeuropa/bcl-theme-default/src/scss/button";
 @import "@openeuropa/bcl-theme-default/src/scss/collapse";
+@import "@openeuropa/bcl-theme-default/src/scss/circular-progress-bar";
```
planctus, line 57: we are not adding this to the sub-themes..?

src/themes/default/src/scss/_circular-progress-bar.scss
```diff
@@ -0,0 +1,147 @@
...
+      transform: rotate(0deg);
+    }
+    100% {
+      -webkit-transform: rotate($degrees);
```
planctus, line 143: same..but let's check the compiled css after removing it..

src/compositions/bcl-project-status/project-status.html.twig
```diff
@@ -28,6 +30,8 @@
 {% set _start_label = start_label|default('') %}
 {% set _end_label = end_label|default('') %}
 {% set _progress = progress|default('') %}
+{% set _chart_progress = chart_progress|default('') %}
```
planctus, line 33: maybe this parameter could be renamed..? I see that technically. let's say, it's a progress value, but we use it as something different, something like "corporate_contribution"..? :)

src/compositions/bcl-project-status/project-status.html.twig
```diff
@@ -28,6 +30,8 @@
...
+{% set _chart_details_list = chart_details_list|default({}) %}
```
planctus, line 34: and this something like `chart_legend`..?

planctus (changes requested):
It's cool, i think we can demo this "template" in the compositions, now.. ;)

src/themes/default/src/scss/_circular-progress-bar.scss
papegaill: there after compilation 👍

src/compositions/bcl-project-status/project-status.html.twig
papegaill: done

src/compositions/bcl-project-status/project-status.story.js
```diff
@@ -0,0 +1,131 @@
...
+        category: "Content",
+      },
+    },
+  };
+};
+
+const applyArgs = (data, args) => Object.assign(data, args);
+
+export default {
+  title: "Features/Project",
```
tibi2303, line 114: Please move it to compositions.

src/themes/default/src/scss/_circular-progress-bar.scss
```diff
@@ -0,0 +1,121 @@
...
+    width: 50%;
+    height: 100%;
+    overflow: hidden;
+    position: absolute;
+    top: 0;
+    z-index: 1;
+  }
+}
+
+.circular-progress-left {
```
tibi2303, line 36: Can you do it the same as on line 59?

tibi2303 (changes requested):
2 small comments, then i think it can be accepted.

PR #256, opened 2022-02-08
feat(subscription): Added subscription page and modal - OEL-1121

tibi2303:


src/compositions/bcl-subscription/subscription.story.js
```diff
@@ -0,0 +1,131 @@
...
+      errorAlert.classList.remove("d-none")
+      form.classList.add('was-validated')
+    </script>
+    `;
+};
+
+export default {
+  title: "Features/Subscription",
+  decorators: [withDesign],
+  parameters: {
```
papegaill, line 113: probably better to put the story in fullscreen like the other features
```suggestion
  parameters: { 
    layout: "fullscreen",
```

src/compositions/bcl-subscription/subscription-block.html.twig
```diff
@@ -0,0 +1,36 @@
...
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+{% set attributes = attributes.addClass(['bg-light', 'px-4-5', 'pt-4', 'pb-4-5']) %}
+
+<div
+  {{ attributes }}
+>
+  <h2 class="fw-normal">{{ _title }}</h2>
```
papegaill, line 25: not sure a `<h2>` tag is needed here, not relevant for SEO. 
```suggestion
  <div class="fs-2 fw-normal">{{ _title }}</div>
```

src/compositions/bcl-subscription/subscription-modal.html.twig
```diff
@@ -0,0 +1,21 @@
+{# Parameters:
+  - form: (form object) (default: {})
+  - success_alert: (alert object) (default: {})
+  - error_alert: (button Object) (default: {})
+  ]
```
papegaill, line 5: `]` can be removed, and maybe specify here the 2 Blocks?
```suggestion
  Blocks:
  - body
  - outer_body
```

papegaill (changes requested):
Well done 👍  I don't really know how the backends will deal with validation, but I imagine that they will find a compromise between JS or Drupal validation on their side. Maybe, we will need to help them on this point.

Looking at the figma, i think there is a missing **subscribe** button in the `bcl-content-banner`.  Not certain that the **print** button should also be included in this feature?

src/compositions/bcl-subscription/subscription.story.js
tibi2303: done.

src/compositions/bcl-subscription/subscription-block.html.twig
tibi2303: done.

src/compositions/bcl-subscription/subscription-modal.html.twig
tibi2303: done. but i guess it can be seen in modal too, it extends it.

papegaill (changes requested):

src/compositions/bcl-subscription/__snapshots__/subscription.test.js.snap
```diff
@@ -0,0 +1,1218 @@
...
+                    >
+                      Category 2
+                    </span>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+          <div
+            class="d-flex justify-content-end"
```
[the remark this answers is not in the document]
tibi2303: We need to update the content_banner fully

papegaill (approved):

src/compositions/bcl-subscription/subscription.story.js
```diff
@@ -0,0 +1,131 @@
+import { withDesign } from "storybook-addon-designs";
+import { correctPaths } from "@openeuropa/bcl-story-utils";
+import header from "@openeuropa/bcl-data-header/data--simple";
+import footer from "@openeuropa/bcl-data-footer/data";
+import content from "@openeuropa/bcl-news/data/data_content.js";
+import banner from "@openeuropa/bcl-content-banner/data/data.js";
+import subscription from "@openeuropa/bcl-subscription/subscription.html.twig";
+import subscriptionData from "@openeuropa/bcl-subscription/data/data.js";
+import sidebar from "@openeuropa/bcl-inpage-navigation/data--simple";
```
planctus, line 9: but with this "data" we don't get a working inpage-navigation in this story...

tibi2303: Done and also I updated the sidebars in Page composition too and deleted some unnecessary imports on news.

src/compositions/bcl-subscription/subscription-block.html.twig
```diff
@@ -0,0 +1,41 @@
+{% spaceless %}
```
planctus, line 1: i like the fact that you created a template for this, it could be potentially used, but i would like the modal inclusion not to be part of this template. Is it going to complicate things..? I also think we should expose these templates we are using in the features, i think you can add a package, with a story for this in compositions.

src/components/bcl-modal/modal.html.twig
```diff
@@ -51,6 +53,7 @@
         {%- block header _header -%}
       </div>
       {% endif %}
+      {%- block outer_body _outer_body -%}
```
planctus, line 56: not sure about the name for this...i agree in adding this "region" in the modal template, but to me that space is really only for alerts, so i would name this block "messages"...

src/compositions/bcl-subscription/subscription-block.html.twig
tibi2303: That was i did on the first commit.. then i changed it.. don't really know why :D I will redo it now

planctus (commented):
Few small things, the demos are really good (can you  disable the controls there.?)

PR #267, opened 2022-02-16
feat(links): Added default underline for anchors - OEL-1180

tibi2303:


planctus (commented):
I'm not really sure we are creating an understandable approach about links with this.
The requirements are not enough detailed and they leave a lot to a case by case evaluation, worst situation for us.
I don't know exactly how to review this pr, what i notice is that:

-  like this we are setting the offset for the underline also for the "text-underline-hover" links and it seems this is only meant for the "generic" link.
- It looks to me that this selector we are using is not well named, to me here we should introduce a semantical distinction between the two "variants", i would finally name "standalone" this selector, because that is the only criteria for evaluating how a link should look like.

src/themes/default/src/scss/base/_utilities.scss
```diff
@@ -62,11 +62,24 @@ $utilities: map-merge(
   )
 );
 
+a {
+  text-underline-offset: 2px;
+}
+
+a:not(.text-underline-hover),
```
planctus, line 69: this looks weird, you can chain the "nots"...

tibi2303: I restructered the scss and told romain about your suggestion. We didn't had a style for visited state, maybe we should.

tibi2303: Talked with romain, we will include this visited ticket next sprint.

PR #273, opened 2022-02-23
feat(carousel): Carousel update - OEL-1197

tibi2303:


src/components/bcl-carousel/carousel.story.js
```diff
@@ -70,6 +72,16 @@ const getArgTypes = (data) => {
...
       },
     },
+    rounded_indicators: {
+      name: "rounded indicators",
+      type: { name: "boolean" },
+      description: "If set to true the indicators would be circles",
+      table: {
+        type: { summary: "boolean" },
+        defaultValue: { summary: "false" },
+        category: "Behaviour",
```
planctus, line 82: we have a style category, i think this should be placed there

src/components/bcl-carousel/carousel.html.twig
```diff
@@ -29,6 +30,7 @@
 {% set _prev_label = prev_label|default('') %}
 {% set _next_label = next_label|default('') %}
 {% set _with_indicators = with_indicators|default(false) %}
+{% set _rounded_indicators = rounded_indicators|default(false) %}
```
planctus, line 33: is this an opt-in thingy, is not the default or the "only" available style of the bcl carousel..? In case it is just close this conversation ;)

papegaill (changes requested):
the dark version of the carousel does not work correctly at the level of styles, caption and indicator don't become negative
Screenshot 2022-03-07 at 17 00 19

src/components/bcl-carousel/carousel.html.twig
tibi2303: default is squared, if you want rounded you set to true

src/components/bcl-carousel/carousel.html.twig
```diff
@@ -29,6 +30,7 @@
 {% set _prev_label = prev_label|default('') %}
 {% set _next_label = next_label|default('') %}
 {% set _with_indicators = with_indicators|default(false) %}
```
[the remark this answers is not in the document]
tibi2303: image
image
image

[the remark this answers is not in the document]
tibi2303: it should be good now, thx for feedback

papegaill (approved):

PR #274, opened 2022-02-24
feat(placeholder): Reset em placeholder styles - OEL-1184

tibi2303:
https://www.drupal.org/project/drupal/issues/3233533
https://www.drupal.org/project/bootstrap_barrio/issues/3228330
https://www.drupal.org/project/vartheme_bs5/issues/3246827

I found better to reset the em.placeholder values than to remove a whole component. As i saw on multiple topics on drupal this placeholder class is added only on the em tag.

### DIFF
--- a/src/components/bcl-alert/__snapshots__/alert.test.js.snap
+++ b/src/components/bcl-alert/__snapshots__/alert.test.js.snap
@@ -21,7 +21,12 @@ exports[`OE - Alert danger renders correctly 1`] = `
       >
         Alert heading
       </h4>
-      A simple alert. check it out!
+      <em
+        class="placeholder"
+      >
+        A simple alert
+      </em>
+      . Check it out!
     </div>
     <button
       aria-label="Close"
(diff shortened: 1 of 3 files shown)


planctus (commented):
First reaction was to ask you if you were seriously proposing to override a drupal selector.
I thought i would have preferred renaming the placeholder component in bootstrap more than doing this, if you really think that it has any value for our users.

Then i read a bit in the drupal issues and maybe i changed my mind, this is kind of grotesque, it would in theory be straighforward to actually remove the placeholder class in drupal, but it's not and the proposed solution is to actually rename the bootstrap component as i initially thought.

I don't know, it feels bad and i think that if we do this i don't want this markup to appear, then we know that it doesn't break but i don't see why demoing an "alien" markup, in any case.

We also need to be sure that this would not be "breaking" more than fixing, do you know all the use cases where this selector would appear in drupal, for instance?

We surely need to warn about this the backend developers and i believe that they would prefer the bootstrap component to be excluded, or renamed, more than taking the risk of overriding this selectors.

tibi2303:
@planctus All the comments are regarding this em.placeholder case... I think, not 100% sure though is the only case. I would like to validate this with francesco too, I will ask him tommorow.

brummbar:
I did a quick search in Drupal and I don't see this class being used together with `em` in other places. The only reference I found was `.exposed-filters .current-filters .placeholder` but no idea on which element it is applied.
For me either resolution is fine, I suggested also to drop the component but we can also just apply this fix for now and if any more issues are caused, we can drop the placeholder component altogether in the future.

planctus (approved):

PR #275, opened 2022-02-25
feat(header): Added support for site name as a link in EC and EU variants - OEL-1221

tibi2303:


src/compositions/bcl-header/header.html.twig
```diff
@@ -10,6 +10,11 @@
         classes: ""
       }
   - site_name (string) (default: "")
+  - project_link (object) (default: {})
+    format: {
+      path: "",,
+      extra_classes: "",
```
planctus, line 16: mmhh..why the extra classes? does this need or expect any "customisation" of the rendered html? I think it doesn't and this would just open room for issues. I think this can be a simple string parameter

src/compositions/bcl-header/header.html.twig
```diff
@@ -20,6 +21,7 @@
 {% set _variant = variant|default('neutral') %}
 {% set _navbar = navbar|default({}) %}
 {% set _project_logo = project_logo|default({}) %}
+{% set _project_link = project_link|default({}) %}
```
planctus, line 24: you kept it as an object here

src/compositions/bcl-header/header.html.twig
```diff
@@ -55,23 +57,31 @@
...
   <div class="{{ _project_classes  }}">
     <div class="container">
-    {% if _project_logo is not empty %}
-      <img
-      {% if _project_logo.path is not empty %}
-        src="{{ _project_logo.path }}"
-      {% endif %}
-      {% if _project_logo.alt is not empty %}
-        alt="{{ _project_logo.alt }}"
+    {% if _project_link is defined and _project_link is not empty %}
```
planctus, line 60: it's an internal variable that you define, so no need to check if it's defined

src/compositions/bcl-header/header.html.twig
```diff
@@ -55,23 +57,31 @@
...
+        {{- _site_name -}}
+        </span>
       {% endif %}
-      />
-    {% endif %}
-    {% if _site_name is not empty %}
-      <div class="{{ _site_name_classes }}">
-       {{- _site_name -}}
-       </div>
+    {% if _project_link is defined and _project_link is not empty %}
```
planctus, line 83: same here

src/compositions/bcl-header/header.html.twig
tibi2303: done.

src/compositions/bcl-header/header.html.twig
tibi2303: done.

src/compositions/bcl-header/header.html.twig
tibi2303: done.

src/compositions/bcl-header/header.html.twig
tibi2303: done.

PR #276, opened 2022-02-28
feat(icons): Added custom icons to all themes - OEL-1222

tibi2303:


src/themes/default/bcl-builder.config.js
```diff
@@ -91,6 +93,16 @@ module.exports = {
       },
     },
   ],
+  sprite: [
+    {
+      entry: path.resolve(nodeModules, "bootstrap-icons/icons/"),
+      dest: path.resolve(outputFolder, "icons/"),
+      options: {
+        file: "bootstrap-icons.svg",
```
planctus, line 101: still thinking about this file name... The script is generating a "bcl-icons.svg" if you don't pass the file name, i think it's better, the bootstrap-icons.svg will stay as an available version provided by bs with all the icons, this sprite will be something namespaced, i see that in the example we use "bcl-default-icons.svg" and i think that in the end is the right name for it.
Now i understand that you will have to find and replace all the occurrencies of this file name but i think it's the right thing to do :)

tibi2303: Done.

src/themes/default/bcl-builder.config.js
```diff
@@ -91,6 +93,15 @@ module.exports = {
       },
     },
   ],
+  sprite: [
+    {
+      entry: path.resolve(nodeModules, "bootstrap-icons/icons/"),
```
planctus, line 98: mmh..i don't like this, we wrap bootstrap everytime in our bcl-bootstrap package.
At the moment in that package we are only copying the sprite from bootstrap-icons, we can copy also all the svgs and then retrieve those from our package.

src/themes/joinup/bcl-builder.config.js
```diff
@@ -1,3 +1,5 @@
+const iconList = require("../default/src/icons/icons");
```
planctus, line 1: ahhh, no, why using a relative path, retrieve this from the package

src/themes/ucpkn/bcl-builder.config.js
```diff
@@ -1,3 +1,5 @@
+const iconList = require("../default/src/icons/icons");
```
planctus, line 1: same here, from the package

tools/story-utils/index.js
```diff
@@ -268,24 +268,12 @@ export const getIconControls = (type, data) => {
 export const getIconNames = () => {
   return [
     "none",
-    "alarm",
+    "upload",
     "arrow-left",
     "arrow-right",
```
planctus, line 273: it seems we removed many, maybe we can add some more in this filtered list..?

planctus (changes requested):
Pretty good, a few things that can be improved, especially the "extension" of the bcl-bootstrap package including also all the svg sources of the icons.

.gitignore
```diff
@@ -8,6 +8,7 @@ bootstrap/plugins
 bootstrap/js
 bootstrap/scss
 bootstrap/*.svg
+bootstrap/icons/*.svg
```
planctus, line 11: you can write bootstrap/**/*.svg instead of line 11 and 12

PR #283, opened 2022-03-03
feat(language-switcher): Added language switcher - OEL-1243

tibi2303:


src/compositions/bcl-language-switcher/language-switcher.story.js
```diff
@@ -0,0 +1,23 @@
+import { correctPaths } from "@openeuropa/bcl-story-utils";
+import { withDesign } from "storybook-addon-designs";
+import demoData from "@openeuropa/bcl-language-switcher/data/data.js";
+import languageSwitcher from "@openeuropa/bcl-language-switcher/language-switcher.html.twig";
+
```
planctus, line 5: can you add a play function for this when executed in chromatic that opens the dropdown? At the moment it's broken in storybook, there is js error

planctus (changes requested):
Fix the error and let's add some sense fun to the screenshot we take in chromatic :)

src/compositions/bcl-language-switcher/language-switcher.story.js
tibi2303: done.

PR #285, opened 2022-03-03
feat(project): Project feature for ucpkn - OEL-1237

tibi2303:


src/compositions/bcl-project/project.html.twig
```diff
@@ -5,58 +5,108 @@
 {% endblock %}
 
 {% block content %}
-  {% include '@oe-bcl/bcl-project-status/project-status.html.twig' with project_details %}
 
-  <div class="border-bottom mb-4">
-    {% include '@oe-bcl/bcl-description-list/description-list.html.twig' with details_list only %}
-  </div>
+  {% if project_status.main_title is not empty %}
```
planctus, line 9: i see that you decided to do this in a different way than the "recommended" one. I'm not really sure there is any value in extending a "feature" template at all, especially if this introduces a more complex data structure that is already overwhelming and if it results in changes about the markup of the other "themes". 
This would mean that the project would need to be updated also in the oe-bootstrap-theme while it seems to me that this could be avoided, regardless of the extension of the current project template or its override in the ucpk theme. 

I believe we should not distribute these templates, they are clearly only for an internal use, it's just confusing to have them in the distributed theme package, need to see how to do this.

src/compositions/bcl-project/__snapshots__/project.test.js.snap
```diff
@@ -2494,8 +2490,8 @@ exports[`OE - Project page (closed) renders correctly 1`] = `
             id="summary"
           >
             
-        Summary
-      
+          Summary
```
planctus, line 2493: it seems we introduced some wthitespaces issues here. But this test should not have any change in my opinion, we didn't have to touch the "project" demo for the default theme in any way. It seems to me that it contains changes that would need to be applied on the default project implementation in drupal, which was not foreseen and should be avoided.

src/themes/ucpkn/src/data/project/data.js
```diff
@@ -0,0 +1,494 @@
+const drupalAttribute = require("drupal-attribute");
```
planctus, line 1: Mmh..we never introduced a data folder in the themes, but i see that it makes probably sense in this case.
But we have to npmignore these files so that we don't ship them in the theme package, at the same time this is an issue because there is no package on npm containing the needed src files for the "package" feature to work on ucpkn. 

So, basically i think these data files should be released in a package and this package can either be:

-  a data package
-  data files contained in a "feature" package

I think it's still ok to ship these files in the "feature" npm package, but i would also be fine with the creation of a dedicated data package.

src/themes/ucpkn/src/stories/project.story.js
```diff
@@ -0,0 +1,141 @@
...
+import header from "@openeuropa/bcl-data-header/data--simple";
+import dataListing from "@openeuropa/bcl-project/data/data_listing";
+import dataContent from "@openeuropa/bcl-theme-ucpkn/src/data/project/data";
+import dataStatus from "@openeuropa/bcl-theme-ucpkn/src/data/project/data_status";
+import dataExtraDetails from "@openeuropa/bcl-theme-ucpkn/src/data/project/data_extra_details";
+import footer from "@openeuropa/bcl-data-footer/data";
+import detailsPage from "@openeuropa/bcl-project/project.html.twig";
+import listingPage from "@openeuropa/bcl-base-templates/listing-page.html.twig";
+
+const social_buttons = `
```
planctus, line 18: we have dozens of data files imported and then we have this directly defined in the story..? :)

planctus (changes requested):
I'm not sure what all the changes in the project test are about, but i wouldn't expect changes there in the "default" test, i would have instead expected a separated test for ucpkn specifically.

src/themes/ucpkn/src/stories/project.story.js
tibi2303: done. added directly into data file.

src/themes/ucpkn/src/data/project/data.js
tibi2303: added inside the feature package

src/compositions/bcl-project/__snapshots__/project.test.js.snap
tibi2303: I added a new test for ucpkn but i don't see why there are a lot of changes, as you can inspect it a bit, is a matter of a div removed for description lists and a div added in gallery

tibi2303: also whitespaces fixed.

src/compositions/bcl-project/data/ucpkn/data_extra_details.js
```diff
@@ -0,0 +1,59 @@
...
+      },
+      {
+        label: "Description",
+        path: "#description",
+      },
+      {
+        label: "Partner organisations",
+        path: "#partners",
+      },
+      {
```
planctus, line 45: why these, what are these for? it seems a broken behaviour this way..

src/themes/ucpkn/src/stories/project.story.js
```diff
@@ -0,0 +1,117 @@
...
+    layout: "fullscreen",
+    controls: {
+      disable: true,
+    },
+    badges: ["stable"],
+    design: [
+      {
+        name: "Mockup",
+        type: "figma",
+        url: "https://www.figma.com/file/NQlGvTiTXZYN8TwY2Ur5EI/BCL-Features?node-id=4298%3A144965",
```
planctus, line 100: don't we have a figma for this..? It seems this link currently points to version 1.1 of this feature, which is really confusing since we are demoing v1

src/themes/ucpkn/src/stories/project.story.js
```diff
@@ -0,0 +1,117 @@
...
+      },
+    ],
+  },
+};
+
```
planctus, line 105: there are accessiblity issues, can you check if they are relevant and in case you fix them?

planctus (changes requested):
I think we are kind of ok, please just check these few things, for me it's fine, let's see the designer's review

src/compositions/bcl-project/data/ucpkn/data_extra_details.js
tibi2303: done.

src/themes/ucpkn/src/stories/project.story.js
tibi2303: done. There is one that is not relevant.. when i fixed them, romain suggested to keep the heading order as it is.

src/themes/ucpkn/src/stories/project.story.js
tibi2303: done.

planctus (approved):

PR #286, opened 2022-03-07
feat(icon): Added missing icons - OEL-1278

tibi2303:


src/components/bcl-badge/__snapshots__/badge.test.js.snap
```diff
@@ -124,7 +124,7 @@ exports[`OE - badge renders correctly dismissible 1`] = `
         class="bi icon--s"
       >
         <use
-          xlink:href="#x"
+          xlink:href="#x-lg"
```
planctus, line 127: image
This is how it looks like.
This is so incredibly confusing, we include an icon without the size so it gets icon-s by default, we use an x-lg for the icon name and we set height and width to make it micro in this pr...what is exactly happening..? :)

tibi2303: But this is a bug i introduced, i will fix it. thank you

That's the name of the icon.. is not really related to the size but mostly to how long the x stretches
image
image

PR #289, opened 2022-03-09
feat(person): Added person feature - OEL-1268

tibi2303:


src/compositions/bcl-person/.npmignore
```diff
@@ -0,0 +1,2 @@
+__snapshots__
```
planctus, line 1: it's private, no need for the npmignore

src/compositions/bcl-person/data/data--listing.js
```diff
@@ -0,0 +1,201 @@
...
+          {
+            label: "Clear",
+            type: "submit",
+            variant: "light",
+          },
+        ],
+      },
+    },
+  },
+  listing: {
```
planctus, line 48: do we need everytime this huge file of data? Maybe we can dedicate some time to reorganise these data files we use "repeatedly" in many different "features", not in this pr, eh.. :)

tibi2303: I agree, I think in the files for each dedicated feature package should be only the custom data for it, in rest we should "built"
it inside the story, I will take some time today to do that.

src/compositions/bcl-person/project.test.js
```diff
@@ -0,0 +1,47 @@
...
+  test("listing renders correctly", () => {
+    expect.assertions(1);
+    return expect(renderListing(demoListing)).resolves.toMatchSnapshot();
+  });
+
+  test("details renders correctly", () => {
+    expect.assertions(1);
+    return expect(renderDetails(demoDetails)).resolves.toMatchSnapshot();
+  });
+});
```
planctus, line 47: it seems you didn't add the package as a dependency for the bcl-twig-templates package.

src/compositions/bcl-person/data/data--listing.js
tibi2303: I think some packages we won't even need a custom data file for it.

src/compositions/bcl-person/.npmignore
tibi2303: done.

src/compositions/bcl-person/project.test.js
tibi2303: done.

planctus:
can you check joinup, there is something in the left sidebar...
image

tibi2303:
@planctus I can't see it..
image

planctus:
this alignment seems approximative also in the default theme, but joinup is more visible

image

Also the border in the design seems closer to the heading than it is in our demo

planctus (approved):

PR #290, opened 2022-03-10
feat(storybook): Reorganising paragraphs  OEL-1311

planctus:


src/compositions/bcl-description-list/description-list.story.js
```diff
@@ -23,3 +23,5 @@ export const Default = () => descriptionList(correctPaths(demoData));
 
 export const Horizontal = () =>
   descriptionList(correctPaths(demoDataHorizontal));
+
+Horizontal.storyName = "Horizontal Description List";
```
tibi2303, line 27: I think this can be Horizontal - removing Description List, it's already inside that..

src/compositions/bcl-links-block/links-block.story.js
```diff
@@ -90,11 +89,6 @@ export default {
 
 export const LinksBlock = (args) => linksBlock(applyArgs(demoData, args));
 
+LinksBlock.storyName = "Links block";
```
tibi2303, line 92: can this be default? it's weird to be the single item without it.

src/compositions/bcl-contact-form/contact-form.story.js
```diff
@@ -54,7 +54,7 @@ const clientValidation = (story) => {
 };
 
 export default {
-  title: "Features/Contact form",
+  title: "Paragraphs/Contact",
```
tibi2303, line 57: I find this more to be a feature, it is a full page layout, the same as the other features... you can resolve it if you don't think so.

src/compositions/bcl-links-block/links-block.story.js
tibi2303: image

tibi2303 (changes requested):
Small and subjective stuff

tibi2303 (approved):

PR #295, opened 2022-03-11
feat(gallery): Added gallery paragraph - OEL-1312

papegaill:


src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -0,0 +1,90 @@
...
+        {% if loop.index % 5 == 0 or loop.last %}
+          </ul>
+        {% endif %}                
+        {% if loop.index > 6 and loop.last %}
+          </div>
+        {% endif %}
+      {% endfor %}
+    {% endif %}
+
+    <div class="modal" id="{{ _modal_id }}" tabindex="-1" aria-labelledby="{{ _modal_id }}" aria-hidden="true" data-bs-backdrop="false">
```
planctus, line 72: this will also be an include, right..?

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -0,0 +1,90 @@
+{% spaceless %}
+
+{#
+Parameters:
```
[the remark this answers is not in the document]
papegaill: Yes sure, I intended to review this, at first I tried to be as close as possible to the parameters for the carousel, 
I reviewed the parameters for the component, now

src/compositions/bcl-gallery/gallery.html.twig
papegaill: as I have additional elements on this modal, such as the pager etc. I did not opt ​​for an include in this case, I have the impression that I will have to deal only with the blocks for this and it seems a little over kill?

src/compositions/bcl-gallery/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "name": "@openeuropa/bcl-gallery",
+  "author": "European Commission",
+  "license": "EUPL-1.2",
+  "version": "0.20.0",
+  "description": "OE - BCL gallery",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@openeuropa/bcl-link": "^0.20.0"
```
planctus, line 11: we have icon and carousel.. ;)

planctus (commented):
We started creating always a data folder also in these compositions, you migh also want to npm ignore it, once you create it

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -0,0 +1,167 @@
...
+
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+{% set attributes = attributes.addClass('bcl-gallery') %}
+
+{# Print the result #}
+
+{% if _items is not empty%}
```
planctus, line 49: we miss a space..ah, this made me remember that we also have twigcs in the lib...

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -0,0 +1,167 @@
...
+{% set _id = id|default('gallery-' ~ random(1000)) %}
+{% set _title = title|default('') %}
+{% set _title_tag = title_tag|default('h2') %}
+{% set _counter = counter|default('') %}
+{% set _toggle_collapsed = toggle_collapsed|default('') %}
+{% set _toggle_expanded = toggle_expanded|default('') %}
+{% set _next_label = next_label|default('Next') %}
+{% set _prev_label = prev_label|default('Previous') %}
+{% set _items = items|default([]) %}
+{% set _icon_path = icon_path|default({}) %}
```
planctus, line 35: this is a string, no..?

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -0,0 +1,167 @@
...
+          <div class="bcl-gallery__thumbnails">
+        {% endif %}
+        {% if loop.index == 6 %}
+          <div class="bcl-gallery__thumbnails-collaspe collapse" id="{{ _collpase_id }}">
+        {% endif %}
+        {% if loop.index % 6 == 0 or loop.first %}
+          <ul class="bcl-gallery__grid">
+        {% endif %}
+
+        <li>
```
planctus, line 80: can we try to respect the indentation for the html tags here..?

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -0,0 +1,167 @@
...
+            {{ _counter|replace({'%d': _items|length}) }}
+          {%- endif -%}
+        </div>
+      </div>
+
+      {% for _item in _items %}
+        {% if loop.first %}
+          <div class="bcl-gallery__thumbnails">
+        {% endif %}
+        {% if loop.index == 6 %}
```
planctus, line 73: i think we can make this configurable, no..? maybe with 6 as default

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -0,0 +1,167 @@
...
+          
+          <div class="modal-body d-flex align-items-center justify-content-center">
+            {% include '@oe-bcl/bcl-carousel/carousel.html.twig' with {
+              id: _carousel_id,
+              with_controls: true,
+              with_indicators: false,
+              prev_label: _prev_label,
+              next_label: _next_label,
+              autoplay: false,
+              items: items,
```
planctus, line 157: pass the local variable here

src/compositions/bcl-gallery/gallery.story.js
```diff
@@ -0,0 +1,20 @@
+import { withDesign } from "storybook-addon-designs";
```
planctus, line 1: We should take care also of the chromatic snapshots for this. You can check in existing stories what we do, it's basically adding .chromatic-ignore to all the images that are loaded "randomly" so that the diff will not report any change about the images.
In a few cases a story file has been added to integrate the stories we normally render with specific ones meant for visual regression tests.
If you name this file test-gallery.story.js and you run yarn start dev it will be loaded, start dev reproduces chromatic so you can use that to "develop" this story.
What i would expect is a story where we would click on a thumbnail so that we also take a snapshots of the modal open with the carousel.

PR #298, opened 2022-03-16
feat(content-banner): Content banner update - OEL-1323

tibi2303:


src/compositions/bcl-content-banner/content-banner.story.js
```diff
@@ -7,9 +7,10 @@ import contentBanner from "@openeuropa/bcl-content-banner/content-banner.html.tw
 
 const button = { ...dataDefault.button };
 
-const getArgs = (data) => {
+const getArgs = () => {
   return {
     background: "gray",
+    image_size: "lg",
```
planctus, line 13: we set this as md by default in the template, why do we demo it with this "lg" that makes it look bad..? :)

src/compositions/bcl-content-banner/content-banner.html.twig
```diff
@@ -95,29 +91,23 @@
 <div class="{{ _css_classes|join(' ') }}">
   <div class="container">
     <div class="row">
```
planctus, line 93: i think this is a leftover, i guess you introduced this when adding this button the first time and it creates a wrong markup, the row here doesn't make sense because the columns are in a children node and indeed they are wrapped into a "row" container. So, this "row" shouldn't simply be here, i think

src/compositions/bcl-content-banner/content-banner.story.js
```diff
@@ -26,14 +27,24 @@ const getArgTypes = () => {
...
-      name: "Action button",
+      name: "action button",
       type: { name: "boolean" },
       description: "Toggle button",
       table: {
         type: { summary: "object" },
         defaultValue: { summary: "{}" },
       },
     },
+    image_size: {
```
planctus, line 38: Maybe now that we added this, we should take three snapshots in chromatic.
For doing this you can create a test-content-banner.story.js where you only add the two stories for the two image sizes not rendered in the default story, You can also render the button in chromatic so that we have it in the snapshots

planctus (commented):
It seems good to me, although we are not yet really using it, let's prepare some additional story for chromatic to render the banner with different "image sizes".

src/compositions/bcl-content-banner/content-banner.story.js
tibi2303: tested something and forgot to make it md

planctus (approved):

PR #302, opened 2022-03-17
feat(pagination): Add first and last params and fix icon alignment - OEL-1316

papegaill:


src/components/bcl-pagination/pagination.story.js
```diff
@@ -62,6 +63,20 @@ const getArgTypes = (data) => {
       },
     },
   };
+  if (data.last && data.first) {
```
planctus, line 66: if you don't want this to appear in the "with custom icons" story you also need to do this in getArgs()
But, i'm a bit confused, there is no harcoded icon name in the template, so in this situation i wouldn't see the reason why showing a story using different icons, there is no "default"...
But i think there shouldn't be the need to choose the icons when using the template unless you want to set custom ones (is this also foreseeen by the deisgners..? cause otherwise we could simply hardcode those icons names

src/data/pagination/data.js
```diff
@@ -1,6 +1,11 @@
 module.exports = {
   aria_label: "Test aria label",
-  enable_icon: false,
+  enable_prev_next_icon: false,
+  enable_first_last_icon: true,
```
planctus, line 4: i wonder why having this as booleans. Then we still look for the icon to be defined in the object, so it seems to me that we could just have an icon boolean parameter in these objects (first last next prev) and basing on that we would build the icon object needed without the need for these "toggles".

src/themes/default/src/scss/_pagination.scss
```diff
@@ -0,0 +1,8 @@
+.pagination {
```
planctus, line 1: Do we really really need this..? There is nothing we can do with markup to get this..?

planctus (commented):
I like it, it's very flexible, but i think we should futher simplify the implementation by simply expecting a boolean in each of the object to identify whether to set an icon or use text and we can build ourselves the icon objects (to be seen if the icons can be customised or not but i don't neessarily see the reason why).

We can also reduce the stories to just the default one, see the related comment.

For the css, i would really try to avoid adding custom scss fiels for such small "tweaks" but if it's really needed then ok, what can we do...

src/themes/default/src/scss/_pagination.scss
papegaill: yes to properly align the icons vertically

src/data/pagination/data.js
papegaill: needed as we have the default text icons (as fallback)

planctus:
@papegaill, i still believe that we should be able to build ourselves these icons objects instead of expecting data from the user, we have the "toggles" for that, then if the user chose to have icons and we have the icon path (i would add this parameter directly in the component) we set the right icon ourselves. We can still look for an icon object in these items to let the user override them, if there is any need for that.

I would also remove this story for "custom icons", it's confusing and currently buggy,  but i believe we shouldn't promote the usage of custom icons for this component, i'm not even sure it should be possible to use custom icons at all.

PR #304, opened 2022-03-21
feat(date-block): Updated date-block - OEL-1336

tibi2303:


src/compositions/bcl-date-block/date-block.story.js
```diff
@@ -45,6 +48,33 @@ const getArgTypes = () => {
...
+      name: "end year",
+      type: { name: "string" },
+      description: "End Year",
+      table: {
+        type: { summary: "integer" },
+        defaultValue: { summary: "" },
+      },
+    },
   };
 };
```
planctus, line 80: don't we have a file on figma to link this demo to..?

src/compositions/bcl-date-block/test-date-block.story.js
```diff
@@ -0,0 +1,8 @@
+import dateBlock from "@openeuropa/bcl-date-block/date-block.html.twig";
+import demoData from "@openeuropa/bcl-date-block/data/data--end-date";
+
```
planctus, line 3: You will not see this snaphot in chromatic, this commit didn't have the [chromatic] comment, so there will be no build for this.. ;)

src/compositions/bcl-date-block/test-date-block.story.js
```diff
@@ -0,0 +1,8 @@
...
+export default {
+  title: "Compositions/Date block",
+};
+
```
planctus, line 7: image

I haven't found the figma file, i don't exactly know what are the use cases this should support, but it looks to me than this is a common use case and it doesn't look good to me..

src/compositions/bcl-date-block/test-date-block.story.js
planctus: ah, no, because there is no [skip-actions] either, you removed that amending the comment..?

src/compositions/bcl-date-block/date-block.story.js
tibi2303: added.

src/compositions/bcl-date-block/test-date-block.story.js
tibi2303: image

I think it's not a valid use case, it will always have the month available

PR #307, opened 2022-03-23
feat(content-banner): Added links in content banner - OEL-1359

tibi2303:


planctus (changes requested):
I don't understand the reason why you added a story only in chromatic, basically how will the users see that there is this option now available in the content banner..? Please add a control as a toggle for those links in the default story, not showing them by default.

src/compositions/bcl-content-banner/content-banner.story.js
```diff
@@ -6,6 +6,7 @@ import dataDefault from "@openeuropa/bcl-content-banner/data/data";
 import contentBanner from "@openeuropa/bcl-content-banner/content-banner.html.twig";
```
planctus, line 6: We need to add the file extension to have the panel showing "source" code, so import data.js and this will be shown in the "data" panel

tibi2303: added the links on the chromatic stories (small image and large image)

src/components/bcl-card/card.html.twig
```diff
@@ -223,6 +225,31 @@
     {% endfor %}
     </div>
   {%- endif -%}
+  {%- if _links is not empty and _links is iterable -%}
+    <div class="mt-2-5">
```
planctus, line 229: i don't understand why taking more space, then you might have one badge and one link and it doesn't make sense but this is on the design side..

src/components/bcl-card/card.html.twig
```diff
@@ -223,6 +225,31 @@
...
+    {% for _link in _links %}
+      {% if _link.attributes is empty %}
+        {% set _link = _link|merge({
+          attributes: create_attribute()
+        }) %}
+      {% endif %}
+      {% set _link = _link|merge({
+          attributes: _link.attributes.addClass('mb-2')
```
planctus, line 237: am i wrong or you are setting a margin bottom on an inline element here..? :)
I wonder if it wouldn't be better to wrap the element with a span anyhow instead of setting attributes to be passed for a class to appear..

src/compositions/bcl-content-banner/content-banner.story.js
```diff
@@ -56,7 +67,11 @@ const applyArgs = (data, args) => {
     data.action_button = button;
   } else {
     delete data.action_button;
-    data.attributes.removeClass(["col-md-10"]);
+  }
+  if (args.links) {
+    data.links = links;
+  } else {
+    delete data.links;
```
planctus, line 74: by the way it look pretty bad with badges, links and action button..but i guess that it's like this..

src/components/bcl-card/card.html.twig
```diff
@@ -223,6 +225,31 @@
...
+        })
+      %}
+      {% if not loop.last %}
+        {% set _link = _link|merge({
+            attributes: _link.attributes.addClass('me-4-5')
+          })
+        %}
+      {% endif %}
+      {% block card_link %}
```
planctus, line 246: but...why is this a block..? how would they use this inside a loop..? if you want to make it a twig block this has to include all the loop..i see that the same has been done for the badges but i don't really understand how this would serve any scope..

planctus (changes requested):
We need to fix the markup and the style of the links and i don't get these two blocks we are adding inside a loop, let's fix it also for the badges that was probably already in place..

src/compositions/bcl-content-banner/test-content-banner.story.js
```diff
@@ -1,17 +1,20 @@
 import { correctPaths } from "@openeuropa/bcl-story-utils";
 import dataDefault from "@openeuropa/bcl-content-banner/data/data";
+import dataLinks from "@openeuropa/bcl-content-banner/data/data--links";
 import contentBanner from "@openeuropa/bcl-content-banner/content-banner.html.twig";
 
+const demoData = { ...dataDefault, ...dataLinks };
+
 export default {
   title: "Compositions/Content banner",
```
planctus, line 9: didn't you mention the possibility to add also the action button, why you didn't do it, then..? :)

tibi2303: Added the action button in a separate file and included it the same as the links.

src/components/bcl-card/card.html.twig
tibi2303: redone that

src/components/bcl-card/card.html.twig
planctus: But..the idea of setting a block for a content that involves the include of a component is not wrong, potentially this could be overridden in order for them to use their own include, what was wrong was defining that block inside a loop. So, put the blocks back but include the whole loops.

src/components/bcl-card/card.html.twig
```diff
@@ -223,6 +225,31 @@
...
+          attributes: create_attribute()
+        }) %}
+      {% endif %}
+      {% set _link = _link|merge({
+          attributes: _link.attributes.addClass('mb-2')
+        })
+      %}
+      {% if not loop.last %}
+        {% set _link = _link|merge({
+            attributes: _link.attributes.addClass('me-4-5')
```
planctus, line 234: i still don't see the reason why not wrapping these links and badges with whatever, a span,  instead of checking for attributes, creating them if not in place and adding a class to an element, it seems an overhead to me.

src/components/bcl-card/card.html.twig
tibi2303: indeed, i readded then properly and also repositioned the action_button after the service_links, now we can support both inside content_banner

planctus (approved):

PR #317, opened 2022-04-04
feat(alert): Styles udpates - OEL-1391

tibi2303:


src/themes/default/src/scss/base/_utilities.scss
```diff
@@ -63,13 +63,10 @@ $utilities: map-merge(
   )
 );
 
+/* stylelint-disable no-descending-specificity */
 a {
   text-underline-offset: 2px;
 
-  &:visited {
```
tibi2303, line 69: They are not taking effect if they styles for visited hover are set on top, so i needed to put them at the bottom but the lint was giving the error and i needed to disable it.

src/themes/default/src/scss/_alert.scss
```diff
@@ -0,0 +1,10 @@
+.alert {
+  .alert-content {
+    color: $dark;
+  }
+}
+@each $state, $value in $theme-colors {
```
planctus, line 6: why do we need this..? My assumption was that we would have just been using the bs mixin, only needing to change the value of a few vars..

tibi2303: in figma, the text color is based on the variant... no other way

src/themes/default/src/scss/_alert.scss
```diff
@@ -0,0 +1,10 @@
+.alert {
```
planctus, line 1: In my idea, we would not have needed a custom scss for applying these changes, isn't this something we could do with a utility..?

src/themes/default/src/scss/_alert.scss
tibi2303: i will do the approach as you mentioned below, i will add the icon color in template directly.

src/themes/default/src/scss/_alert.scss
tibi2303: indeed, thank you, i will change this

src/themes/default/src/scss/_alert.scss
planctus: i'm not sure, isn't the icon getting the theme color..? couldn't you pass text-{themeColor} class to the icon instead?

planctus (approved):

PR #318, opened 2022-04-05
feat(project): Added main contributor - OEL-1399

tibi2303:


src/compositions/bcl-project/project.html.twig
```diff
@@ -48,9 +48,32 @@
...
+  {% if chart_lists.main_title is not empty %}
     <h3
       class="fw-bold my-4"
+      {% if chart_lists.main_title_id is not empty %}
+      id="{{ chart_lists.main_title_id }}"
+      {% endif %}
+    >{{- chart_lists.main_title -}}</h3>
+  {% endif %}
+  {% if chart_lists.listing is defined and chart_lists.listing is not empty %}
+    {% for lists in chart_lists.listing %}
```
planctus, line 64: can't we make a template out of this thing..? It could be used as such

tibi2303: done, also there were some problems on ucpkn, fixed those also.

PR #321, opened 2022-04-07
feat(publication): Added publication - OEL-1398

tibi2303:
[skip chromatic]


planctus:
I guess we can first evaluate the heading pr so that you would be able to set this id for the file passing an attribute from the data as title_attributes

planctus:
I'm also wondering a bit about the possibility of making the "authors" list a separate template, we can maybe ask be if they would benefit from that..

tibi2303:
@planctus done, for the authors, i would still wait for that, we have only one place where we use these authors. On the second time we see them, we make a template.

PR #323, opened 2022-04-08
feat(project-status): Splitting in two indpendent templates - OEL-1426

planctus:


src/compositions/bcl-project-status/project-status.html.twig
```diff
@@ -32,8 +20,17 @@
 {% set _start_label = start_label|default('') %}
 {% set _end_label = end_label|default('') %}
 {% set _progress = progress|default('') %}
-{% set _corporate_contribution = corporate_contribution|default('') %}
-{% set _chart_legend = chart_legend|default({}) %}
+
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
```
donquixote, line 26: I wonder, can we pipe this, and should we?
```twig
{% set attributes = (attributes ?: create_attribute()).addClass([
  'bcl-project-status',
  'border-bottom',
  'pb-4',
  'mb-4'
]) %}
```
(I think both `?:` and `??` would work here)

src/compositions/bcl-project-status/project-status.html.twig
```diff
@@ -32,8 +20,17 @@
...
+
+{% set attributes = attributes.addClass([
+  'bcl-project-status',
+  'border-bottom',
+  'pb-4',
+  'mb-4'
+]) %}
```
donquixote, line 33: A question for this template, but also for others: Can we remove classes through parameters?
For this template the question is especially about the mb-4 and border-bottom.
However I am also curious in general how we could do this. E.g. in the card pattern I also wanted to remove a class that was added by default.

If not, we should only add the minimum classes we need.

https://oelibrary.netlify.app/default/?path=/story/features-project--ongoing
Here the border-bottom is added to all elements in a sequence except for the last one.
If the subsequent sections are all gone, because of optional fields being empty, then the status element should not have the border-bottom.

This means that the decision on whether or not to add the border-bottom has to be done outside, not as part of the template.

For easier control, I am even considering to use a `<hr class="border-bottom opacity-100 mt-0 mb-4"/>` elements between items, so that I don't need an awkward condition to determine if this is the last element (possible in for loop with .last, but not in `|map()` filter, and so I don't need to add classes to elements with different structure.

Another option would be a utility class on the container that contains the list of items to be separated. The utility class could then put a border-bottom on each except the last element. I have done similar tricks in SRB. But I think this is going too far.

If we cannot easily remove classes from outside, I'd say it is better to only add what we don't need to remove later:  `['bcl-project-status', 'pb-4']`. I think these are inherent to the element itself, whereas the `border-bottom` and `mb-4` describe its relation to other elements.

src/compositions/bcl-project-status/project-status.html.twig
```diff
@@ -42,16 +39,7 @@
...
 {% set _bg = _bgs[_status]|default('secondary') %}
 
-
-{% if _title is not empty %}
-  {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
-    title: _title,
-    title_tag: _title_tag,
-    attributes: _title_attributes,
-  } only %}
-{% endif %}
```
donquixote, line 52: Good :)

src/compositions/bcl-project-status/project-status-contributions.html.twig
```diff
@@ -0,0 +1,56 @@
+{% spaceless %}
```
donquixote, line 1: Should this be called 'project-status-contributions` or just `project-contributions`?
Your choice.

src/compositions/bcl-project-status/project-status.html.twig
donquixote: We could do
```twig
{% set attributes = (attributes ?: create_attribute()).addClass([
  'bcl-project-status',
  'border-bottom',
  'pb-4',
  'mb-4'
]|filter(class => class not in classes_to_remove)) %}
```
But this would pollute the parameter signature. We would have to add `anti_attributes` or `classes_to_remove`. It seems like complexity creep.

donquixote: _or_ we could have a `.last-child-no-border` utility class :)

```css
.last-child-margin-none > *:last-child {
  border-bottom: none !important;
}
```

src/compositions/bcl-project-status/project-status.html.twig
planctus: eh, but this would have a different result, what we want to do is to leave the possibility of adding attributes, not overriding the default ones needed for styling of the component ;)

src/compositions/bcl-project-status/project-status-contributions.html.twig
planctus: we can rename it, sure..

src/compositions/bcl-project-status/project-status.html.twig
planctus: The answer to the question is that whenever it's possible we try to avoid writing custom css and stick to bs markup.
It's a choice, i wouldn't change the approach, but at the same time it is possible, obviously, to handle different use cases in different ways and this is already happening, when there is no easy way to setup a screen with bs markup only, custom css is written, but as a "last resource".

It seems to me that you are reporting two isues that basically go back to one:
if we would have border bottom for all the elements in the design of this "content type page" there would be no issue, so i'm going to ask the deisgners if it's acceptable to have that border also on the last item and we close this discussion for this particular use case.

Whenever you feel that markup shouldn't be set by default on our side you can raise an hand but you shouldn't simply try to override it yourself, if there is some needed improvements, tweaks, changes in a template for you to able to use it, you should ask and let us review your request and act on the template if we "accept" your requested change.

emeryro: on design side, we can add a border bottom in this case, that's not a big deal.
If we have this kind of issue again later, we may have to think at a more global approach, but for now it can be solved easily

src/compositions/bcl-project-status/project-status.html.twig
planctus: no, it would not have different results, but it really looks confusing although it is a more concise expression.
I will keep it as it is for readability and because this kind of approach of creating attrs when they are empty is a standard, default thing we do for all the templates the same way. Thanks for proposing an interesting solution, code wise, in any case ;)

src/themes/ucpkn/src/stories/project.story.js
```diff
@@ -82,6 +93,9 @@ const plannedDemo = {
 const demoListing = {
   ...baseData,
   ...dataListing,
+  pagination: pagination,
+  filter_button: filterButton,
+  sort_select: sortSelect,
```
tibi2303, line 98: I did the same stuff in my pr, can you just add the project status data files too? In order to not have them created in the story but imported from the project-status story directly
image
image

src/compositions/bcl-project-status/project-contributions.html.twig
```diff
@@ -0,0 +1,56 @@
...
+  - corporate_contributions (string) (default: '') // value every 5 percent
+  - legend (boolean or object (default: false)
+  - chart (boolean) (default: false)
+
+  Blocks - 
+    project_status_contrib_chart
+    project_status_contrib_legend
+#}
+
+{% set _corporate_contributions = corporate_contributions|default('') %}
```
tibi2303, line 13: I think it would be better for this to have a default, the one we are using now for project.

src/themes/ucpkn/src/stories/project.story.js
tibi2303: sure

tibi2303 (approved):

PR #325, opened 2022-04-10
feat(project): Added main contributor - OEL-1399

tibi2303:


planctus (commented):
i've just noticed that we miss the badge for the project version.. ;)

tibi2303:
@planctus added 1.1 for project, all the version badges should match the figma ones.

PR #326, opened 2022-04-11
chore(storybook): Replacing problematic plugin for html source

planctus:


.storybook/main.js
```diff
@@ -4,11 +4,11 @@ const stories = ["../src/*/*/*.story.js"];
 
 const addons = [
   "@storybook/addon-docs",
+'@openeuropa/storybook-addon-code/register'
   "@storybook/addon-controls",
   "storybook-addon-designs",
   "@storybook/addon-viewport",
-  "@storybook/addon-a11y",
-  "@whitespace/storybook-addon-html",
+  "@storybook/addon-a11y",,
```
tibi2303, line 11: same here

tibi2303 (changes requested):
Thank you for this, seems like we don't have the js errors and the badges are back.

3 small stuff...

tibi2303 (approved):

PR #328, opened 2022-04-12
feat(headers): Added subthemes header - OEL-1437

tibi2303:


planctus:
It seems all good to me and now the features in ucpkn they show what they are supposed to show, at least regarding hte header, do you know anything about their footer..? Is it really the "neutral" one that we are currently showing..? Do you know anything about it..? :)

For joinup i wonder if we couldn't propose them to put their customisations in BCL, it would be nice to demo the real joinup header and footer, but it's not up to us

There is one story, though, that should also have this applied to, which is "multilingual", there i don't see the ucpkn header.

Good job!

tibi2303:
For the multilingual this are the showcases in figma (Neutral / EC / EU) https://www.figma.com/file/NQlGvTiTXZYN8TwY2Ur5EI/BCL-Features?node-id=5356%3A215100

I am not sure it's needed to show the theme header there.

I asked a question on ux channel for the custom footers, waiting for respone.

PR #331, opened 2022-04-18
feat(page-titles): Added titles for features - OEL-1466

tibi2303:


src/compositions/bcl-event/event.story.js
```diff
@@ -72,6 +76,12 @@ const demoListing = {
     link: false,
     title: false,
   },
+  banner: {
+    ...pageTitleBanner,
+    title: {
```
planctus, line 81: i was surprised to see this structure, now i wonder...we recently aligned all the titles using the new template and standard parameters like title, title_tag, title_attributes. It seems that the card was left behind, there we kept the title as an object and this is creating this ambiguity. Did we do that for backward compatibility..? I'd like to align everything, because i don't see why keeping this different sructure for then using the same template. We can evaluate whether to maintain support for a title object in the card, but i believe we should add the standard parameters in order to support the new template. What do you think..?

PR #332, opened 2022-04-19
feat(table): Added responsive story for tables - OEL-1474

tibi2303:
[skip chromatic]

### DIFF
--- a/src/data/table/data--responsive.js
+++ b/src/data/table/data--responsive.js
@@ -1,0 +1,249 @@
+module.exports = {
+  table_head: {
+    cells: [
+      {
+        content: "#",
+      },
+      {
+        content: "Heading",
+      },
+      {
+        content: "Heading",
+      },
+      {
+        content: "Heading",
+      },
+      {
+        content: "Heading",
+      },
+      {
+        content: "Heading",
+      },
+      {
+        content: "Heading",
+      },
+      {
+        content: "Heading",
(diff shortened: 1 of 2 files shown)


src/data/table/data--responsive.js
```diff
@@ -0,0 +1,249 @@
+module.exports = {
+  table_head: {
```
planctus, line 2: are we sure that we need a separated story for demoing this? And even if we do, couldn't we build a table with realistic data? Also the choice of using a tablet viewport doesn't help, we are demoing responsiveness we should use "responsive" also in the viewport so that you can test the different breakpoints. But to me,  this is not a variant, is just an option and could be demoed via controls in the default story, no..?

tibi2303: done, i added more content to the default story and removed the responsive one.

planctus: eh, ok, but now we have the control with a default that is different than the one in the template. I don't even see the reason why not setting it by default to be responsive in the template, but if we don't want to change that, then the control would be better set as empty, as it is when using the template.

tibi2303: added empty as default for table_responsive

tibi2303: i think the default table should not have the responsive enabled.

planctus (approved):

PR #335, opened 2022-04-25
feat(content-banner): Using single parameter for the action bar in content banner - OEL-1517

planctus:


tibi2303 (changes requested):
image
The icons are not visible, anything else is good.

tibi2303 (approved):

PR #336, opened 2022-04-25
feat(headings):  Update heading hierarchy - OEL-1495

papegaill:


tibi2303 (commented):
This PR is related more to https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1495
I guess you can continue up with this ticket,
What i was reffering in OEL-1496 is that the card.html.twig does not use the new heading template and this card template has a lot of occurances in other templates. In short, changing will impact all the features.
Anyway, this is a good start for OEL-1495, you need to also change the heading tags based on this rules (
-sections in the main area are H2
-sections in the sidebar are H3
-elements inside components do not use headings (card title, fact and figures data, ...). They use fixed font values
).

tibi2303 (commented):
The sidebar title should be h3 everywhere - the title inside bcl-page-navigation is h5.

PR #338, opened 2022-04-26
feat(spacings): Added general spacings for content banner and footer - OEL-1507

tibi2303:


papegaill (approved):

src/compositions/bcl-base-templates/content-type.html.twig
```diff
@@ -116,16 +116,16 @@
       {% block related %}
         {{ related }}
       {% endblock %}
+      {% block share %}
```
tibi2303, line 119: @papegaill i am not sure yet if this a breaking change or not, seems not but should be checked
@planctus if you want to have a look over too, that would be great.

PR #340, opened 2022-04-27
feat(card): Changed title to heading - OEL-1496

tibi2303:


src/components/bcl-card/card.html.twig
```diff
@@ -171,16 +170,15 @@
...
-      {% set _title_attributes = title.attributes ?: create_attribute() %}
-      {% set _title_attributes = _title_attributes.addClass("card-title") %}
-      {{ _title_attributes }}
-    >
-      {{- title.content -}}
-    </{{ _title_tag }}>
-  {%- endif -%}
+  {% if _title is not empty %}
+    {% set _title_attributes = _title_attributes ?: create_attribute() %}
+    {% set _title_attributes = _title_attributes.addClass("card-title") %}
```
planctus, line 175: single quotes :)

src/components/bcl-card/card.html.twig
```diff
@@ -1,12 +1,9 @@
...
 {# Parameters:
-  - title (object) (default: {})
-    format: {
-      content: '',
-      tag: '',
-      classes: ''
-    }
+  - title: (string) (default: '')
+  - title_tag: (string) (default: 'h5')
+  - title_attributes (drupal attrs)
```
planctus, line 6: but..we should have all the parameters from the bcl-headings, including the title_link, i see that in the listings examples you didn't use it and the links are passed as strings in the title parameter. Although this would work, even in the implementation, is not the recommended way, using the title_link will enable all the features that are available through the link component basically and we should "promote" this usage of the new bcl-headings template.

src/compositions/bcl-subscription/__snapshots__/subscription.test.js.snap
```diff
@@ -256,7 +256,7 @@ exports[`OE - Subscription renders correctly 1`] = `
             class="breadcrumb-item"
           >
             <a
-              href="/example"
+              href="/example.html"
```
planctus, line 259: wow, how does it come that we had all these broken links..

planctus (changes requested):
Thanks a lot, suprised to see all the links you had to fix. It's good but i think we should fully benefit from the bcl-heading and use the title_link in the card as well, not "suggesting" a shortcut like passing the <a> tag in the title parameter, which is supposed to be "plain text", although we don't handle and don't validate the data input.

src/components/bcl-card/card.html.twig
tibi2303: done. I changed the data to use title_link. I find it a bit weird but it's a good approach that the label is not needed and actually the title param is set as the label, might be confusing

src/components/bcl-card/card.html.twig
tibi2303: done, thanks

src/compositions/bcl-content-banner/content-banner.html.twig
```diff
@@ -71,10 +71,9 @@
 } %}
 
 {% set _card = {
-  title: title|default({})|merge({
-    tag: title.tag ?: 'h2',
-    attributes: create_attribute().addClass(['fw-bold']),
-  }),
+  title: title,
+  title_tag: title_tag ?: 'h2',
+  title_attributes: create_attribute().addClass(['fw-bold']),
```
planctus, line 76: maybe we should support also passing attributes here, like we do in the other templates.
I guess it's correct, instead, not exposing the title_link parameter, because this content banner is basically going to hold the page title.

src/compositions/bcl-fact-figures/fact-figures.html.twig
```diff
@@ -2,30 +2,31 @@
 
 {#
   Parameters:
-  - "title" (string) (default: '')
-  - "title_tag" (string) (default: 'h2')
-  - "variant" (string) (default: '') options:
+  - title (string) (default: '')
+  - title_tag (string) (default: 'h2')
+  - title_link: (link object) (default: {})
```
planctus, line 7: mmhh..here i don't think we needed the title_link, but anyway..

src/compositions/bcl-file/file.html.twig
```diff
@@ -38,6 +39,7 @@
 {% set _icon = icon|default({}) %}
 {% set _title = title|default('') %}
 {% set _title_tag = title_tag|default('h2') %}
+{% set _title_link = title_link|default({}) %}
```
planctus, line 42: Also here, i don't see why creating a title for a file download "paragraph" that is a link, but if you want to say that we always use the same parameters then add it also in the content banner

src/compositions/bcl-group/data/data--landing.js
```diff
@@ -420,8 +420,10 @@ module.exports = {
       },
       items: [
         {
-          title: {
-            content: '<a class="standalone" href="/example">Card title</a>',
+          title: "Card title",
+          title_link: {
+            path: "/example.html",
+            standalone: true,
```
planctus, line 426: don't you think that standalone should come by default from the template..?

planctus (changes requested):

src/components/bcl-card/card.html.twig
planctus: this is a fallback, so that we try to build a proper title even in case of a "misaligned" data, but we shouldn't promote this usage

src/compositions/bcl-content-banner/content-banner.html.twig
planctus: since i saw that you applied all the parameters everywhere, do it also here, instead, add also the title_link

tibi2303: done, also added the same approach to content_banner

src/compositions/bcl-file/file.html.twig
tibi2303: now everywhere is the same way.

src/compositions/bcl-group/data/data--landing.js
tibi2303: added in heading

PR #341, opened 2022-04-29
feat(dates): Aligning date format used in the library - OEL-1541

planctus:
Dates and "metas" have been aligned with the mocks when possible, the possibility to use a date block instead of the image has been added for the content banner since in one mock this "solution" was used instead of the standard image.
Some small fixes have been applied to the stories regarding isChromatic || chromatic, there might still be some inconsistency around, though.


src/compositions/bcl-event/event.story.js
```diff
@@ -57,9 +58,14 @@ const demoData = {
...
   ...dataPage,
-  banner: banner,
+  banner: {
+    ...banner,
+    image: {},
+    date,
+  },
   files: files,
 };
+console.log(demoData1);
```
tibi2303, line 68: remove this please

src/compositions/bcl-content-banner/__snapshots__/content-banner.test.js.snap
```diff
@@ -40,7 +40,129 @@ exports[`OE - Content banner renders correctly 1`] = `
                 <span
                   class="text-muted me-3"
                 >
-                  8 hours ago
+                  17 October 2019
+                </span>
+              </div>
+              <p>
+
```
tibi2303, line 47: I saw that you removed the whitespaces for recent activities, can you do the same here, please?

tibi2303 (changes requested):
Small stuff.

tibi2303 (approved):

PR #342, opened 2022-05-02
Fix(bcl-icon): Deprecated XLink URL reference attributes

arturu:
References: 
- https://www.w3.org/TR/SVG/linking.html#XLinkRefAttrs 
- https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

### DIFF
--- a/src/components/bcl-icon/icon.html.twig
+++ b/src/components/bcl-icon/icon.html.twig
@@ -28,7 +28,7 @@
 <svg
   {{ attributes }}
 >
-  <use xlink:href="{{ _path ~ "#" ~ _name }}"/>
+  <use href="{{ _path ~ "#" ~ _name }} xlink:href="{{ _path ~ "#" ~ _name }}"/>
 </svg>
 
 {% endspaceless %}


planctus (commented):
@arturu, not sure where the idea of directly creating a pull request on the repo comes from, your initiative or any kind of procedure i'm not aware of..? It costs time for us to review code not part of the normal flow and i'm not sure it would make sense to act this way more than giving you access to the support channels for BCL where to make a request, report an issue, etc..
In any case, taking into account your request, we are fully aware of the deprecation of the xlink:href attr, but the truth is that this is going on since 5 years and there is no guarantee that href will ever get a wider support than xlink:href currently has, so i don't see the reason why "polluting" our code duplicating this attribute now.
Bootstrap is still promoting the usage of xlink:href in the usage examples for icons:
https://icons.getbootstrap.com/#usage
and in general there is no confidence about this changing in the future either, so...
I will close this pr, you can still push for this creating an issue here on github, starting and participating to a discussion in our support channels, but for how i see this, the "deprecation" can safely be ignored, indeed your proposal is to use both attributes because you probably know that href per se would not get the same support that xlink:href has... ;)

PR #345, opened 2022-05-05
feat(user): Update to v1.1 - OEL-1538

tibi2303:


src/themes/default/src/scss/_banners.scss
```diff
@@ -128,6 +138,12 @@
         margin-left: 0;
       }
     }
+    .bcl-content-banner {
```
tibi2303, line 141: here this selector is inside .container

planctus (approved):

PR #346, opened 2022-05-06
feat(badge): Added hover/focus state - OEL-1560

tibi2303:


src/compositions/bcl-group/__snapshots__/group.test.js.snap
```diff
@@ -4522,9 +4528,11 @@ exports[`OE - Group listing Member renders correctly 1`] = `
                     />
                   </svg>
                 </span>
-              </span>
-              <span
-                class="mb-2 me-2 badge bg-light text-dark"
+              </a>
+              <a
+                class="mb-2 me-2 badge dismissible bg-light text-dark"
+                href="#"
```
planctus, line 4534: can we use /example.html for all these urls..?

src/components/bcl-badge/badge.html.twig
```diff
@@ -27,6 +27,11 @@
   {% set _classes = _classes|merge(['rounded-pill']) %}
 {% endif %}
 
+{% if _dismissible %}
+  {% set _classes = _classes|merge(['dismissible']) %}
+  {% set _url = _url|default('#') %}
```
planctus, line 32: but why this default..?
But why this logic, i'd say..? So now if it's disimissible is always going to be a link with a # as default url, don't think this was the idea..

src/components/bcl-badge/badge.html.twig
```diff
@@ -35,6 +40,8 @@
 
 {% if _background in ['light', 'warning'] %}
   {% set _classes = _classes|merge(['text-dark']) %}
```
planctus, line 42: Considering the amount of css we are producing for these badges it feels this might actually come from the css directly..but ok

src/themes/joinup/src/scss/oe-bcl-joinup.scss
```diff
@@ -84,7 +84,7 @@
 @import "@openeuropa/bcl-theme-default/src/scss/pagination";
 
 // JoinUp Custom styles
-@import "@openeuropa/bcl-theme-joinup/src/scss/badges";
+@import "@openeuropa/bcl-theme-joinup/src/scss/badge";
```
planctus, line 87: do we need to rename this file..? I see you did it for consistency, but this could potentially create an issue on the application side, i wouldn't do it..

planctus (changes requested):

src/components/bcl-badge/badge.html.twig
tibi2303: I am setting this as a url for the focus behaviour (tab navigation), being <span> can't really be used this behaviour. Also i am using preventDefault() in the initBadges util function.
I changed the behaviour of removable action to the whole bagde as i talked with romain because it would be better to have a bigger clickable area.

planctus:
veeeeery nice, so much better, can i ask you to add stories in chromatic to test these statuses you added..?
Hover and focus, then it would be nice also to have one story where you click on the close icon to test the "new" functionality but the resulting screenshot would be entirely blank so it's not probably worth.. :)

planctus (approved):

PR #364, opened 2022-05-25
feat(improvements): Card variant footer/header and multiselect close - OEL-1640

tibi2303:


src/themes/default/src/scss/_multiselect.scss
```diff
@@ -46,8 +46,27 @@ select.multi-select {
       text-align: center;
       white-space: nowrap;
       vertical-align: baseline;
+      padding-right: map-get($spacers, "3-5");
```
planctus, line 49: i think we should review the existing css we wrote, when we use bootstrap for setting margins and paddings we kind of support RTL, when we write css we use padding-right, padding-left...this should be padding-inline-end but maybe let's do this in a separated ticket

src/themes/default/src/scss/_card.scss
```diff
@@ -0,0 +1,8 @@
+@each $color, $value in $theme-colors {
```
planctus, line 1: it seems we are only taking care of the shaded colors of  header and footer here, but we should remove the border in all these variants, seeing how they look like in figma, and we have a problem, because we should also adapt the color of the text in a few cases, but this can't happen at the template level, i think, because these are theme colors and in some sub-theme there might be different combinations needed...

planctus (changes requested):
Left a couple of comments..

src/themes/default/src/scss/_card.scss
planctus: mmhh...i wrote that the solution should not be coming from the template, we are requesting additional data to the user introducing potential risks of weird combinations and additionally breaking backward compatibility. I would rather do this via css, you can set the default values in the default theme, if any override is currently needed this can happen in the sub-theme's css. But there are also the badges to be taken care of, For those we can just update the demo, using a different style for them via the markup or removing those badges in any variant is chosen, up to you..

planctus (approved):

PR #371, opened 2022-06-08
feat(icons-list): Adding a page displaying all the available icons - OEL-1684

tibi2303:


planctus:
Cool!! But can we align those icons..? It looks a bit messy right now :)

src/compositions/bcl-colors/colors.story.js
```diff
@@ -126,7 +126,7 @@ const applyArgs = (data, args) => {
 };
 
 export default {
-  title: "Colors",
+  title: "Resources/Colors",
   decorators: [withCode],
```
planctus, line 130: i don't think we need this panel here.. ;)

src/compositions/bcl-icons-list/icons-list.story.js
```diff
@@ -0,0 +1,19 @@
+import { correctPaths } from "@openeuropa/bcl-story-utils";
+import icons_list from "@openeuropa/bcl-icons-list/icons-list.html.twig";
+import icons from "@openeuropa/bcl-theme-default/src/icons/icons.js";
+import defaultSprite from "@openeuropa/bcl-theme-default/icons/bcl-default-icons.svg";
+
+let list = icons.map((icon) => icon.substring(0, icon.length - 4));
+
+console.log(list);
```
planctus, line 8: i stil have to understand the reason why the linter is not complaining when we forget a console.log().. :)

src/compositions/bcl-icons-list/icons-list.story.js
```diff
@@ -0,0 +1,19 @@
...
+export default {
+  title: "Resources/Icons",
+  parameters: {
+    layout: "fullscreen",
+    controls: { disable: true },
+  },
+};
+
+export const Default = () =>
+  icons_list(correctPaths({ icons: list, path: defaultSprite }));
```
planctus, line 19: this confuses me a bit, why do we need to use correctPaths() if we are importing the sprite, or viceversa, why do we import the sprite if we set the right path through the utility function..?

planctus (changes requested):
A few things and we are ok

tibi2303:
@planctus done, thx

planctus (approved):

PR #372, opened 2022-06-09
feat(title): Title harmonized - OEL-1685

tibi2303:


planctus (changes requested):
mmhh..wasn't this supposed to "improve" somehow the current situation..? From what i see all the headings are going to have the fw-medium class associated, therefore it seems unreasonable to expect this from the user data, the heading template could set this by default.
But i also believe that due to the requirements, there was some misunderstanding, we changed into paragraphs also "headings" in the "content" area when there is no sidebar (user), which doesn't seem to meet the requirements.
Furthermore i don't understand the reason why content in the left sidebar should not be considered content and because of this "loosing" the headings, which then are going to be missing in mobile where they will be at the top of the "content" area. This is more about the requirements, per se, i started a discussion on this with our deisgners.

planctus:
So, as agreed with the designers titles in the sidebar will be h2s but styled as they currently are (smaller font size). This should happen via css so that we don't need user data to "distinguish" them.
At the end these are the requirements:

- no accessibility error concerning the heading hierarchy
- main content area contains heading starting with h2
- titles visually smaller and not changing based on heading hierarchy for "aside" content (sidebar, share, feedback...)

planctus (approved):

PR #375, opened 2022-06-10
feat(events): Adding time information,  aligning with figma mockup, adding controls - OEL-1695

planctus:
Quite a mess:

1) While reproducing the example in figma with the title only in the content banner i realised that in the template we were misusing attributes, applying them to the internal card object instead of the root element of the template and i adapted the content banner template in order to use the main attributes object to set attributes for the component itself.

2) This was then used to add a padding bottom to the content banner which is only set by the template when there is the grey background, not entirely sure why.

3) I find per se pretty useless to align everytime a chaging "content" in the mocks, that's what it is, content, and i wouldn't care at all, but since this was mainly a request about updating "strings", then i went aligning as much as possible with the "content" we currently have in figma.

4) To avoid meaningless complaints about the fact that the stories are less than the amount of pages in figma to demo all the different use cases, with only a title, with a text, with an image, whatever, i added controls in the listing story to be able to render these different use cases.

5) There are small tweaks on the scripts side, yarn lint is going to run all the linting processes, including prettier.

6) I noticed that the margins we were setting in the content of the banner for the "date", aka a random string were a bit exaggerated, this produced many changes in the screenshots using the content banner, thus the amount of file changed.


src/compositions/bcl-event/event.story.js
```diff
@@ -44,17 +60,54 @@ export default {
...
+  }
+
+  return data;
+};
+
 export const FullPage = () => event(correctPaths(demoData));
 
 FullPage.storyName = "Event page";
 FullPage.decorators = [initScrollspy];
+FullPage.parameters = {
```
papegaill, line 91: Strange link color in the footer of this story:
Screenshot 2022-06-13 at 13 24 30

src/compositions/bcl-event/event.story.js
```diff
@@ -44,17 +60,54 @@ export default {
...
+  controls: {
+    disable: true,
+  },
+};
 
 export const FullPage1 = () => event(correctPaths(demoDateData));
 
 FullPage1.storyName = "Event page with date";
 FullPage1.decorators = [initScrollspy];
+FullPage1.parameters = {
```
papegaill, line 101: Screenshot 2022-06-13 at 13 28 10

- date-block date should be 17 OCT - 21 OCT in this case
- date-block width is larger in figma 160px, here we have 120px
- spacing between date-block and card content is to wide compare to figma, date-block seems lost

figma:
Screenshot 2022-06-13 at 13 33 05

papegaill, line 101: - on figma "Register now" button is full-width on mobile
- label is "Register" insteed of "Register now"

src/compositions/bcl-event/event.story.js
papegaill: Same for the language and log-in buttons in the header

papegaill (changes requested):

planctus:
thanks @papegaill, for being just an update on a single feature this involved many changes here and there, which are hopefully improvements, i checked for "regressions" myself but in any case i will make a build on chromatic to see the results in the screenshots diff.

papegaill (approved):

PR #378, opened 2022-06-14
feat(inpage): Flat list, removing scrollspy stories - OEL-1731

planctus:


tibi2303:
It's indeed better that you removed these demos, but i see a bug that is not concerning this pr, can you fix the margin bottom of the title?
image
a mb-0 on the title should fix it

tibi2303 (approved):

PR #380, opened 2022-06-17
feat(glossary): Adding glossary listing and detail - OEL-1696

planctus:


tools/story-utils/index.js
```diff
@@ -368,6 +368,8 @@ export const correctPaths = (data) => {
         data[prop] = data[prop].replaceAll("/icons.svg", iconPath);
       } else if (data[prop].includes("flags.svg")) {
         data[prop] = flagPath;
+      } else if (data[prop].includes("/example.html")) {
+        data[prop] = `/example.html#${Math.random().toString(36).slice(2, 7)}`;
```
tibi2303, line 372: image
this goes to a 404

src/compositions/bcl-glossary/glossary-listing.html.twig
```diff
@@ -0,0 +1,45 @@
...
+
+{% block content_row %}
+	<div class="col-lg-4 col-md-6 col-12 mb-5">
+  	{% include '@oe-bcl/bcl-search-form/search-form.html.twig' with search only %}
+  </div>
+		{% include '@oe-bcl/bcl-pagination/pagination.html.twig' with glossary only %}
+	<div class="col-12">
+	  <div class="d-md-flex justify-content-end mb-md-5 align-items-center border-bottom-md border-top-md border-gray-300 py-5">
+	    {% include '@oe-bcl/bcl-select/select.html.twig' with {
+			  label: "Sort by",
```
tibi2303, line 11: I think it's better for this to come from the data file, not directly in the template, same goes for the other select. They are kind of big, that's the only reason.

tools/story-utils/index.js
planctus: eh, wow, this is not because of this update, it's like this also in production, need to check what we are deploying..

planctus: eh, well..we were only copying the index.html, this has been like this for centuries, i guess..

tibi2303 (approved):

PR #382, opened 2022-06-17
feat(links): Visited links update - OEL-1737

planctus:
It seemed easier to do, i had doubts this could be done at some point. instead.
The specifications regarding the styles that can be associated to a :visited link do not seem entirely consistent between different browser, their "privacy" policy it's basically preventing fancy style changes.
It seems it is possible to fulfill the requirements, though, and i hope the current styles are doing that.
There was a flaw i think in the definition of the "default" links, the :not were not chained and therefore the styles were also being applied to the standalone variant.


tibi2303:
It should be removed also for page-link class i guess
image

planctus:
This cannot be done. simply as that: https://codepen.io/planctus/pen/zYRQjWp

tibi2303 (approved):

PR #385, opened 2022-06-22
feat(inpage): Removing full_page parameter, building example in the story file - OEL-1743

planctus:


src/compositions/bcl-inpage-navigation/inpage-navigation.story.js
```diff
@@ -1,13 +1,52 @@
...
+        ${heading({
+          title: "Heading 3",
+          attributes: new drupalAttribute().setAttribute("id", "item-3"),
+        })}
+        <p>${lorem.generateParagraphs(5)}</p>
+        ${heading({
+          title: "Heading 4",
+          attributes: new drupalAttribute().setAttribute("id", "item-4"),
+        })}
+        <p>${lorem.generateParagraphs(3)}</p>
```
tibi2303, line 40: It is nice that you added this package, but i find that adding this in data file directly would be better. Just thinking of all the story files it would need to modify and I think it will pollute these files with all this content. Not sure if it's possible but i think it would be nice.

planctus: not in this case, i'd say, you will see that in all the other stories the dummy text will be generated in the data file as you suggest, but since this is totally aside of the data sent to the template, i think it's properly defined in the story file where it is used

PR #386, opened 2022-06-22
feat(card-refactoring): Redone usage of card and layouts of content-banner and listings - OEL-1689

tibi2303:
[skip chromatic]


src/themes/default/src/scss/base/_layout.scss
```diff
@@ -1,5 +1,25 @@
-@include media-breakpoint-up(lg) {
```
planctus, line 1: but why did you remove this..?

PR #388, opened 2022-06-23
chore(dummy-text): Adding utility, replacing in every story - OEL-1744

planctus:


src/components/bcl-card-layout/__snapshots__/card-layout.test.js.snap
```diff
@@ -126,9 +128,13 @@ exports[`OE - Card grid renders correctly 1`] = `
...
             class="card-text mb-2"
-          >
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+          />
+          <p>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Lorem ipsum dolor sit amet, consectetur adipiscing elit. 
+            <br />
+             Lorem ipsum dolor sit amet, consectetur adipiscing elit. 
           </p>
+          <p />
```
tibi2303, line 137: It's adding some closed p tags.

planctus: Yep, it was by default wrapping in a <p> tag, but in many cases there was already a paragraph in place, so i made it by default plain text.

tibi2303 (approved):

tibi2303 (changes requested):
At code level, it's ok, but the design look weird now.
image

Also in some place we don't have text at all
image

tibi2303 (approved):

PR #393, opened 2022-06-28
feat(description-list): Added bordered param to description-list - OEL-1777

tibi2303:


src/compositions/bcl-description-list/description-list.html.twig
```diff
@@ -31,8 +33,21 @@
   {% set attributes = create_attribute() %}
```
planctus, line 33: Can we always apply a namespaced selector on the root element of every template..? bcl-description-list, in this case..?

src/compositions/bcl-description-list/description-list.html.twig
```diff
@@ -44,21 +59,32 @@
...
+      >
+      {% if _item.term is iterable %}
+        <dt>
         {% for _term in _item.term %}
-          <dt
-          {% if _variant == 'horizontal' and _term.icon is not empty %}
-            class="d-flex"
-          {% endif %}
-          >
+            {% if _item.term.length > 1 %}
```
planctus, line 81: is this working..? i know about a length filter in twig, not a "method"...

src/compositions/bcl-description-list/__snapshots__/description-list.test.js.snap
```diff
@@ -41,114 +19,294 @@ exports[`OE - Description List horizontal renders correctly 1`] = `
...
+      <dd>
+        Description text goes here.
+      </dd>
+    </dl>
+    <dl
+      class="d-md-grid grid-3-9 mb-2-5"
+    >
+      <dt>
+        <div>
+
```
planctus, line 34: why this empty line..?

tibi2303: image

Cause of the bordered style, i needed to refactor the containers, but still every list has it's own dl tag, and all of them are wrapped inside this div container.

src/compositions/bcl-description-list/description-list.html.twig
tibi2303: done, should a ticket be created for all of these cases?

planctus: i think so.. :)

planctus:
can you just check the indentation in the template..? it's a bit off, as usual try to get html tags correctly indented, twig code is less important..

tibi2303:
@planctus done.

planctus (approved):

PR #398, opened 2022-06-30
feat(namespace-selector): Added root class on multiple templates - OEL-1794

tibi2303:


src/compositions/bcl-links-block/links-block.html.twig
```diff
@@ -31,6 +31,8 @@
   {% set attributes = attributes.addClass(['bg-light', 'px-4', 'py-3', 'rounded-2']) %}
 {% endif %}
 
+{% set attributes = attributes.addClass(['bcl-links-block']) %}
```
planctus, line 34: can we just put this on top so that it is the first class appearing..?

planctus (approved):

src/compositions/bcl-links-block/links-block.html.twig
tibi2303: done

PR #400, opened 2022-07-04
feat(listing): Listing image size parameter and cleanup - OEL-1810

tibi2303:


src/compositions/bcl-listing/data/listing--profiles.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   variant: "default-1-col",
   remove_horizontal_spacer: true,
+  image_size: "small",
```
tibi2303, line 4: After a talked with romain, the images of profiles inside listings should be 120px X 120px. In order to do that I add a new param inside listings called image_size, that is changed only inside profiles - the default is 160px X 160px.

planctus: yeah, but this has also affected the "date" listing where the left column size should be 120px

image

tibi2303: thanks, i changed for dates also.

planctus (approved):

PR #410, opened 2022-07-18
feat(z-index): Lowering values not to conflict with webtools, fixing templates - OEL-1767

planctus:
…


tibi2303:
https://getbootstrap.com/docs/5.0/layout/z-index/
Based on this, I think we may not have conflicts and it's good to be merged.

PR #415, opened 2022-07-29
Make the button classes overridable

doxigo:
Hey there, 
So I noticed that you really can't add any extra classes if you need to do so, I think that would come in handy time to time


tibi2303:
That variable is internal, you should not set classes through that variable but through drupalAttributes if you want extra ones.

doxigo:
Thanks for the update tibi

PR #416, opened 2022-07-31
feat(cleanup): Removed extension from import of js files - OEL-1907

tibi2303:


planctus:
Not sure the reason why you did this, the reason why of the usage of the extension in the file import is related to the "data" panel, basically if we use the extension the content of the file will be visible in the panel otherwise it won't. Now, this makes sense for components, basically, where we import a file containing all the data, it doesn't make much sense when our data is instead the composition of many data files, with or without an "inxex" file grouping them. Basically in these cases it would be better to keep the "data" panel empty.
If we want to do this consistently, ok for me, you can do this in this pr or recreate it from scratch

planctus (approved):

PR #418, opened 2022-08-02
feat(bootstrap-5.2): Aligning components - OEL-1916

tibi2303:


src/components/bcl-offcanvas/offcanvas.html.twig
```diff
@@ -14,6 +14,8 @@
   - extra_classes_body (string) (default: '')
   - extra_classes_header (string) (default: '')
   - extra_classes_close (string) (default: '')
+  - responsiveness (string) (default: '')
+    options: ['sm', 'md', 'lg']
```
planctus, line 18: but there are more options, actually.. :)
.offcanvas-sm
.offcanvas-md
.offcanvas-lg
.offcanvas-xl
.offcanvas-xxl

planctus:
It seems mostly good to me, the UI tests revealed a "leftover" feature: publications listing, about the offcanvas update. The form is not visible. 
But, about this change, I like the fact that we could remove a custom scss, thanks to this bs update, but i think we should possibly make this a backward compatible change, what i mean is that i would try not to break websites currently using "offcanvas".
But how to do this..? Maybe setting by default offcanvas-lg in the template..?

planctus (commented):
It seems mostly good to me, the UI tests revealed a "leftover" feature: publications listing, about the offcanvas update. The form is not visible. 
But, about this change, I like the fact that we could remove a custom scss, thanks to this bs update, but i think we should possibly make this a backward compatible change, what i mean is that i would try not to break websites currently using "offcanvas".
But how to do this..? Maybe setting by default offcanvas-lg in the template..?

src/compositions/bcl-base-templates/sidebar-search.html.twig
```diff
@@ -1,7 +1,16 @@
 {% extends "@oe-bcl/bcl-offcanvas/offcanvas.html.twig" %}
 
+{% set responsiveness = 'lg' %}
+{% set extra_classes_header = extra_classes_header ~ ' d-lg-block' %}
```
tibi2303, line 4: @planctus I set here the responsive breakpoint. I am not sure though if they are using this template for the sidebar. Still setting the default param to lg would break the offcanvas story and the component.

tibi2303: Also seems the title **Filter Options** were missing, so i am setting a class to show them now

planctus: this template for sure not, but most likely they are not even including the offcanvas template itself. The thing is that we have to mark this as a breaking change, then...

tibi2303: if they are not using even the offcanvas template then we have no other way than to set it as breaking change.

tibi2303: still we can keep the previous scss just for the backwards compatibility and let the sidebar as it is in this pr.

planctus (approved):

PR #420, opened 2022-08-04
feat(featured-image): Improvement on width image - OEL-1920

tibi2303:


src/themes/default/src/scss/_featured-media.scss
```diff
@@ -0,0 +1,9 @@
+.bcl-featured-image {
```
planctus, line 1: question: isn't it this something we could achieve with utilities instead of introducing a new css file..?
Surely we don't have an utility for min-content but we could add that.

tibi2303: indeed, i think it's better, i pushed it, made the classes abstract to be reusable

planctus (approved):

PR #423, opened 2022-08-08
feat(heading): Adding stories for heading(s) - OEL-1929

planctus:


tibi2303 (approved):

tibi2303:
@planctus i see that for example the icon-list is a private package and don't have .npmignore. Should it be the same here?

planctus:
the bcl-heading package is not private..

tibi2303:
image
On default i get a white page...

PR #426, opened 2022-08-10
feat(content-banner): Fix container for banner inside tab panel - OEL-1935

tibi2303:


src/themes/default/src/scss/_featured-media.scss
```diff
@@ -0,0 +1,9 @@
+.bcl-featured-image {
```
planctus, line 1: How is this related..?

tibi2303: weird.. i don't know why is this here... deleted

planctus (approved):

PR #434, opened 2022-08-18
chore: collapsed margins

tibi2303:


src/themes/default/src/scss/mixins/_collapsed-margins.scss
```diff
@@ -0,0 +1,30 @@
+@mixin collapsed-margins-pseudo-element($margin) {
```
planctus, line 1: why this name, tiberiu..? It's misleading, not that i super liked the previous naming, but it was more descriptive of what these mixins are really supposed to do.

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,45 @@
+$bcl-spacing-m: 1.5rem;
+$bcl-spacing-l: 2rem;
+$bcl-spacing-xl: 3rem;
+
+.bcl-section {
+  @include clearfix;
+
+  margin-bottom: $bcl-spacing-l;
+  margin-top: $bcl-spacing-l;
```
planctus, line 9: i think we could try to switch from margin-collapse to a mono-directional approach, like using only margin-bottom..

src/themes/default/src/scss/mixins/_collapsed-margins.scss
planctus: i would also put this file in the base folder, no reasons why creating another folder.

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,45 @@
...
+
+  > h2,
```
planctus, line 11: i think we also have to take care of other headings, we don't have requirements so it's not important the "value" we use, but just to make it clear that this would replace the idea of a css for the wysiwyg

src/themes/default/src/scss/_collapsed-margins.scss
```diff
@@ -0,0 +1,23 @@
+// Utility classes to absorb/neutralize vertical margins of child elements.
+// This allows to fully control vertical space using padding, without additional
+// space caused by margins of e.g. <h2> or <p> elements inside the container.
+// This is more reliably than anything with :first-child or :last-child, because
+// it considers only visible block-level elements.
+
```
planctus, line 6: shouldn't this go into the utilities file we already have..?

src/themes/default/src/scss/_collapsed-margins.scss
```diff
@@ -0,0 +1,23 @@
...
+// Neutralizes top margin of first child, and bottom margin of last child.
+.collapsed-margins,
+.children-collapsed-margins > * {
+  @include collapsed-margins();
+}
+
+// Neutralizes top margin of the first block-level child element.
+.eat-margin-top,
```
planctus, line 14: is this a leftover, not renamed..?

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,45 @@
...
+    margin-top: $bcl-spacing-m;
+    &:first-child {
+      margin-top: 0;
+    }
+    &:last-child {
+      margin-bottom: 0;
+    }
+  }
+
+  p:last-child {
```
planctus, line 23: isn't this already covered by the following  > *:first-child?
Actually also the "exceptions" added for headings seems redundant, the rule aforementioned is going to take care about them.

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,45 @@
...
+  > *:first-child {
+    margin-top: 0;
+  }
+
+  > *:last-child {
+    margin-bottom: 0;
+  }
+}
+
+.row + .row .bcl-section {
```
planctus, line 36: i would remove this, it is taken from the europa styleguide where it's probably tailored to some existing use case, in our scenario it seems useless or potentially harmful at the moment.

src/compositions/bcl-section/section.story.js
```diff
@@ -0,0 +1,10 @@
+import section from "@openeuropa/bcl-section/section.html";
+
+export default {
+  title: "Features/Section",
```
planctus, line 4: mmhh..why in "features"..? Maybe resources, or let's consider it a "component"

src/compositions/bcl-project/data/data--extra-details.js
```diff
@@ -44,7 +44,8 @@ module.exports = {
       title: "Summary",
       content: `
       <p>
-        This is the first item's accordion body. It is hidden by default, until the collapse plugin adds the appropriate classes that we use to style each element. These classes control the overall appearance, as well as the showing and hiding via CSS transitions.
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis fermentum facilisis ex, ac porta massa. Lorem ipsum dolor sit amet, consectetur adipiscing elit.
```
planctus, line 47: oh my god, why do we still have lorem ipsum text in our data..?

src/themes/default/src/scss/mixins/_collapsed-margins.scss
planctus: reset-margins as a base name, maybe..? ;)

src/compositions/bcl-base-templates/content-type.html.twig
```diff
@@ -29,11 +30,18 @@
 {% set _with_footer = footer|default(false) %}
 {% set _with_banner = banner|default(false) %}
 {% set _with_sidebar = sidebar|default(false) %}
+{% set _reset_container_margins = reset_container_margins|default(false) %}
 {% set _sidebar_size = sidebar_size|default('normal') %}
 {% set _landing = landing|default(false) %}
 {% set _content_type = content_type|default('') %}
 {% set _classes = ['bcl-node-type'] %}
 
+{% set _container_classes = 'container' %}
```
planctus, line 39: To a certain extent, if we want to consider this a viable solution, this would be a pull request to be merged at some point, to me it's mybe better to apply the sections everywhere removing all the usage of unneded utilities, more than re-creating the same situation of the first poc to make project work without breaking the other demoes.

src/compositions/bcl-base-templates/content-type.html.twig
```diff
@@ -29,11 +30,18 @@
 {% set _with_footer = footer|default(false) %}
 {% set _with_banner = banner|default(false) %}
 {% set _with_sidebar = sidebar|default(false) %}
+{% set _reset_container_margins = reset_container_margins|default(false) %}
```
planctus, line 33: To a certain extent, if we want to consider this a viable solution, this would be a pull request to be merged at some point, to me it's mybe better to apply the sections everywhere removing all the usage of unneded utilities, more than re-creating the same situation of the first poc to make project work without breaking the other demoes.

tibi2303: then in this case, i need to clean all those extra classes everywhere...

planctus: eh, but that is basically what a "structural" spacing approach is meant for.. :)
Let's work with the idea that we are going to merge this pull request once we are ok with it

tibi2303: ok, then i will create a ticket and have this ready for chromatic at the end.

PR #436, opened 2022-08-18
feat(vertical-spacing): General approach for content spacing - OEL-1955

tibi2303:
In this pr:
- Management of vertical spacing ( all content inside pages are inside a div with class .bcl-section that take cares of the spacing between texts and headings and also components - banners, listings )
- Common root namespace for all the components that use headings in order for the default spacing between heading and content


src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,37 @@
...
+.bcl-section {
+  @include clearfix;
+
+  margin-bottom: $bcl-spacing-l;
+  margin-top: $bcl-spacing-l;
+
+  > h2,
+  > .bcl-heading {
+    margin-bottom: $bcl-spacing-m;
+    margin-top: $bcl-spacing-m;
```
tibi2303, line 14: image
For the container class names to remain clean, we can't go for a mono directional approach.

tibi2303: Most i can do is to have this margin-top only on the .bcl-section:first-child, in order to get the spacing.

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,59 @@
...
+
+  > *:last-child {
+    margin-bottom: 0;
+  }
+
+  &:first-child {
+    margin-top: $bcl-spacing-l;
+  }
+
+  > .bcl-listing,
```
planctus, line 44: this is the section component css, why would we style bcl-banner and bcl-listing..? can't we wrap those "components" into a section to get the desiderd spacing..?

src/components/bcl-accordion/accordion.html.twig
```diff
@@ -5,7 +5,7 @@
   - title (string) (default: '')
   - title_tag (string) (default: 'h2')
   - title_link: (link object) (default: {})
-  - title_attributes: (drupal attrs)
+  - title_attributes: (drupal attrs) (default: classes: "mb-4")
```
planctus, line 8: it seems we are still confused about these headings, i'll try to explain the way i see this:
We have an include of the bcl-heading in all the paragraphs templates. 
**These headings should get their style from the section css, .bcl-section h(n), generic enough to trigger all the headings regardless of the fact that they are direct children of the section wrapper**.
This class that you add here then it would override this default style, this would only be justified in case we have a specific requirement, which we don't have, so i expect these utilities to be removed, basically.

Then there might be other headings as part of the component markup, which is the case here for the accordion, if we do like i propose this will then require to reset the default margins on all the other headings, this would be handled adding a utility class to the internal headings (the buttons).

Furthermore in the stories where we demo the paragraphs, including the title, in the story file the mb-4 should be added in order to ensure some spacing for the sake of the demo.

src/compositions/bcl-glossary/glossary-detail.html.twig
```diff
@@ -28,16 +28,15 @@
 	{% include  '@oe-bcl/bcl-heading/heading.html.twig' with {
 		title: 'File Downloads',
 		title_tag: 'h2',
-		attributes: create_attribute().addClass('mt-5', 'mb-4'),
 	} only %}
 
 	{% for file in files %}
 		{% if loop.last %}
```
planctus, line 34: we should try to avoid this, potentially we could use the utility we added, for them it seems difficult to add a class to all the items in the array but the last one, so we have to approach this differently

src/compositions/bcl-group/group-landing.html.twig
```diff
@@ -74,13 +74,19 @@
...
+        title: recent.title,
+        title_tag: 'h2',
+      } only %}
       {% include '@oe-bcl/bcl-recent-activities/recent-activities.html.twig' with recent.block_data only %}
     </div>
     <div class="col-md-4">
-      <h2 class="mt-4 mt-md-0 mb-4 pb-2">{{ members.title}}</h2>
+      {% include  '@oe-bcl/bcl-heading/heading.html.twig' with {
+        title: members.title,
+        title_tag: 'h2',
```
planctus, line 88: All these h2(s) can be removed, it's the default in the heading template.

src/compositions/bcl-group/package.json
```diff
@@ -16,6 +16,7 @@
     "@openeuropa/bcl-listing": "^0.26.0",
     "@openeuropa/bcl-navigation": "^0.26.0",
     "@openeuropa/bcl-recent-activities": "^0.26.0",
+    "@openeuropa/bcl-heading": "^0.26.0",
```
planctus, line 19: not so important but can we keep an alphabetical order..?

src/compositions/bcl-event/event.html.twig
```diff
@@ -5,20 +5,28 @@
...
-  <h2 id="item-2" class="my-4">Related documents</h2>
-  {% include '@oe-bcl/bcl-file/file.html.twig' with files[0] only %}
-  {% include '@oe-bcl/bcl-file/file.html.twig' with files[1] only %}
+  <div class="bcl-section">
+    {% include '@oe-bcl/bcl-description-list/description-list.html.twig' with details_list only %}
+    {% include '@oe-bcl/bcl-heading/heading.html.twig' with {
+      title: 'Description',
+      attributes: create_attribute().setAttribute('id', 'item-1'),
+    } only %}
+    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi sit amet eleifend tortor. In facilisis eros vitae turpis ullamcorper,
```
planctus, line 14: utility for generating random text has been added also to twig as a function "get_dummy_text()" with the same params as the js utility.

src/compositions/bcl-glossary/glossary-detail.html.twig
```diff
@@ -1,43 +1,39 @@
...
-	a dapibus libero porta nec. Sed facilisis dictum vestibulum. Fusce commodo hendrerit diam, pretium tempus
-	leo varius sit amet. Etiam interdum, orci at sagittis luctus, lorem libero tempus mauris, a fermentum libero
-	orci semper lacus. Duis tristique fringilla magna, eu egestas dolor molestie non. </p>
-	<p>Lorem ipsum dolor
-	sit amet, consectetur adipiscing elit. Suspendisse volutpat ultricies massa, a dapibus libero porta nec.
-	Sed facilisis dictum vestibulum.
-	Fusce commodo hendrerit diam, pretium tempus leo varius sit amet. Etiam interdum, orci at sagittis luctus, 
-	lorem libero tempus mauris, a fermentum libero orci semper lacus. Duis tristique fringilla magna,
-	eu egestas dolor molestie non. </p>
+		<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse
```
planctus, line 9: also here, you can use the utility

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,64 @@
...
+
+  > *:first-child {
+    margin-top: 0;
+  }
+
+  > *:last-child {
+    margin-bottom: 0;
+  }
+
+  &:first-child {
```
planctus, line 51: i don't think this makes really sense..i don't understand the logic and indeed the demo is broken, showing a doubled margin. If we are in a margin-bottom only approach, then the upper section is going to create the spacing for the one below, regardless of the fact that they are siblings. This that you did would only if the sections are all siblings and then margin-collapse would take care of "eatring" the doubled spacing. Applying this in the "real world", the content banner would be wrapped in a seciton, this will create the spacing with the left sidebar and the main content, no need to have an internal spacing in the section first item, then...

tibi2303: indeed, i removed it and added the banner inside a section and it seems ok

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,60 @@
...
+  .h2,
+  .h3,
+  .h4,
+  .bcl-heading {
+    margin-bottom: $bcl-spacing-m;
+    margin-top: $bcl-spacing-m;
+    &:first-child {
+      margin-top: 0;
+    }
+    &:last-child {
```
planctus, line 24: This breaks, i mean, it would make sense only if the heading is the last element of the section, but like this we are actually resetting the margin-bottom of the last h2, most likely this is wrong, considering that there is "content" below it.

tibi2303: i think it may indeed happen if the content is not wrapped, in any case i removed it.

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,57 @@
...
+  h3,
+  h4,
+  .h1,
+  .h2,
+  .h3,
+  .h4,
+  .bcl-heading {
+    margin-bottom: $bcl-spacing-m;
+    margin-top: $bcl-spacing-m;
+    &:first-child {
```
planctus, line 21: i think that what you want to have here is :first-of-type more than first:child, you want to reset the margin of the first heading, first-child is only going to target the heading when it is the first element in the section...

src/compositions/bcl-contact-form/contact-form.html.twig
```diff
@@ -1,5 +1,7 @@
 {% extends '@oe-bcl/bcl-base-templates/content-type.html.twig' %}
 
 {% block content %}
-  {% include '@oe-bcl/bcl-form/form.html.twig' with form only %}
+  <div class="bcl-section">
```
planctus, line 4: let's use <section>, instead of a generic div

src/themes/default/src/scss/_section.scss
planctus: You only changed this for the minor headings, but the same applies to the h1, h2 headings, the margin-bottom should not be reset for the last-child and first-of-type should be used instead of first-child

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,54 @@
...
+  > *:first-child {
+    margin-top: 0;
+  }
+
+  > *:last-child {
+    margin-bottom: 0;
+  }
+}
+
+@include media-breakpoint-down(md) {
```
planctus, line 46: we have a problem in mobile, using the section in the left sidebar is creating an empty area due to the margins. Then i don't understand the values you used, it is just a demo, but why did you set a bigger margin in mobile than we have in desktop..? If we want to demo this "feature", we would probably do the opposite, no..?

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,54 @@
+$bcl-spacing-s: 1rem;
+$bcl-spacing-m: 1.5rem;
+$bcl-spacing-l: 2rem;
+$bcl-spacing-xl: 3rem;
+
+.bcl-section {
+  @include clearfix;
+  margin-bottom: $bcl-spacing-l;
+
+  h1,
```
planctus, line 10: the fact that you put also h1 here is adding margin to the "page title" in the content banner, which is not necessarily bad, but then we use the my in the "subtitle" and now it is not needed anymore, can be simply mb-

src/themes/default/src/scss/_section.scss
```diff
@@ -0,0 +1,54 @@
...
+  > *:last-child {
+    margin-bottom: 0;
+  }
+}
+
+@include media-breakpoint-down(md) {
+  .bcl-section {
+    margin-bottom: $bcl-spacing-xl;
+
+    &:first-of-type {
```
planctus, line 50: but then why are we adding this margin here..?

tibi2303: Indeed, by removing this, it should also fix that big spacing from the sidebar

src/themes/default/src/scss/_section.scss
tibi2303: I lowered the spacing to 1rem.

src/themes/default/src/scss/_section.scss
tibi2303: All of the muted text under the titles should have now mb-

tibi2303: still i am thinking now that the spacing is not equal

src/themes/default/src/scss/_section.scss
tibi2303: removed

tibi2303:
I modified the margins of the footer to align them with figma on compositions.
I changed the file aproach - an extra margin class was added from a bad drupal attributes class adition.

PR #447, opened 2022-09-08
feat(description-list): Description List refactor - OEL-2014

tibi2303:


src/compositions/bcl-description-list/data/data--horizontal.js
```diff
@@ -10,7 +10,9 @@ module.exports = {
           path: "/icons.svg",
         },
       },
-      definition: "Description text goes here.",
+      definition: {
```
planctus, line 13: did you agree with backend about this change..? it's always confusing, we had a request to make the label size configurable and we went for a full refactoring here..

src/compositions/bcl-description-list/description-list.html.twig
```diff
@@ -68,120 +64,56 @@
...
-            <span class="p-2 me-2 d-inline-block rounded-1 {{ _term.color }}"></span>{{- _term.label -}}
-          </div>
-        {%- else -%}
-          {{- _term.label -}}
-        {%- endif -%}
-        {% if _item.term|length > 1 %}
-        </div>
+  >
+    {% if _variant == 'horizontal' %}
+    <dl class="row mb-0">
```
brummbar, line 80: Like this we are still rendering multiple `<dl>`, but we should have only one that wraps the all the component.

src/compositions/bcl-description-list/description-list.html.twig
```diff
@@ -36,25 +36,21 @@
...
-  {% endif %}
+  {% set _container_tag = 'div' %}
 {% endif %}
 
 {% if _bordered %}
-  {% set _item_classes = _item_classes ~ ' border-bottom' %}
+  {% set _item_classes = 'border-bottom pb-2-5 mb-2-5' %}
+{% endif %}
+
+{% if _col_size is not empty %}
```
brummbar, line 52: In the call we said to allow to pass classes, so it leaves more freedom to the user. They could define the behaviour different based on viewport, so this just leaves too many combinations possible.

brummbar (changes requested):

src/compositions/bcl-description-list/description-list.html.twig
```diff
@@ -68,120 +64,56 @@
...
-          {% if _variant == 'horizontal' %}</div>{% else %}</dd>{% endif %}
-        {% endfor %}
-        {% if _variant == 'horizontal' %}
-        </dd>
+        {% if _item.definition is iterable %}
+          {% for _definition in _item.definition %}
+            {%- include '@oe-bcl/bcl-description-list/description-list-details.html.twig' with _definition only -%}
+          {% endfor %}
+        {% else %}
+          {%- include '@oe-bcl/bcl-description-list/description-list-details.html.twig' with _item.definition only -%}
```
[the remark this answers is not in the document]
tibi2303: the support for single strings is no longer available, this line is about the support of array of object or single object.
image

src/compositions/bcl-description-list/description-list.html.twig
tibi2303: the markup is no longer correct if we have a dl for all in the horizontal layout.

The problem is that the DL tag supports only one layer inside of DIV tag. But for the column layout we need 2 - row and column.
image

src/compositions/bcl-description-list/description-list.html.twig
tibi2303: indeed, i change it.

src/compositions/bcl-description-list/description-list.html.twig
```diff
@@ -36,25 +36,21 @@
   {% set attributes = create_attribute() %}
 {% endif %}
 
-{% set _item_classes = '' %}
+{% set _item_classes = 'mb-3' %}
+{% set _col_classes = 'col' %}
```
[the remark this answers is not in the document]
tibi2303: this is used for an inner tag, we set the attributes only on the root tag.

src/compositions/bcl-description-list/description-list.html.twig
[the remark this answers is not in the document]
tibi2303: I guess this is again overcomplicated, maybe we should only support arrays of objects and that's all. what do you think?

src/compositions/bcl-description-list/description-list.html.twig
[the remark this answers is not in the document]
tibi2303: also this is no longer available as i changed the parameter to directly add the class. As you said for more possibilities

src/compositions/bcl-description-list/description-list.html.twig
brummbar: Then we are doing things wrong. This was the main point of the remark, if it's a semantic description list it should have only one `<dl>`. I'll guess I'll redo the review with this in mind, but I'm quite disappointed.

tibi2303: @brummbar please have a look, I change to approach more or less the same we had before but without the extra markup. I think it should be good.

src/compositions/bcl-description-list/description-list.html.twig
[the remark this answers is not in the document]
brummbar: > the support for single strings is no longer available
This we can never ever do. We cannot drop support, especially between minors.

In general though, we should always simplify. It should have always supported only one scenario, the array of objects. Or better, we should have supported markup without any additional icon. Users could put the icons by themselves.

brummbar (changes requested):

src/compositions/bcl-description-list/description-list.html.twig
```diff
@@ -6,14 +6,13 @@
   - title_tag: (string) (default: 'h2')
   - title_link: (link object) (default: {})
   - title_attributes (drupal attrs)
-  - bordered (boolean) (default: '')
+  - bordered (boolean) (default: false)
+  - horizontal_size (string) (default: '3-9')
```
brummbar, line 10: In the call we said to drop this and use multiple dl plus normal row/col classes, why it wasn't implemented?

brummbar (changes requested):
I see that now when I have two terms and then one definition, there's extra space between the terms:

image

src/compositions/bcl-description-list/description-list.html.twig
tibi2303: indeed, i was focused on the points in this pr and may have missed that, anyway it's pushed

Also regarding the spacing, it may be the case on your side only
image

I guess this 'new' implementation will not have this bug anymore but tell me if it's still happening.

brummbar:
We have issues on smaller viewports:
image

cc @emeryro

brummbar (approved):

PR #450, opened 2022-09-13
feat(documentation): Updated parameters documentation - OEL-2022

tibi2303:
[skip chromatic]


src/components/bcl-toasts/toasts.html.twig
```diff
@@ -21,15 +21,17 @@
   - attributes (drupal attrs)
 #}
 
-{% set _toasts = toasts|default([]) %}
 {% set _with_wrapper = with_wrapper|default(false) %}
 {% set _wrapper_classes = wrapper_classes|default('') %}
 {% set _wrapper_aria_live = wrapper_aria_live|default('') %}
 {% set _with_container = with_container|default(false) %}
-{% set _container_classes = 'toast-container' %}
+{% set _container_classes = container_classes|default('') %}
```
planctus, line 28: not sure this change really makes sense, you wanted to be consistent using "_classes" but this variable is used for the single toasts wrapper, it is safe since the variable is set in a loop, but it's still confusing.

src/components/bcl-toasts/toasts.html.twig
```diff
@@ -21,15 +21,17 @@
...
+{% set _toasts = toasts|default([]) %}
+
+{% set _classes = 'toast-container' %}
 
-{% if container_classes %}
-  {% set _container_classes = _container_classes ~ ' ' ~ container_classes %}
+{% if _container_classes is not empty %}
+  {% set _classes = _classes ~ ' ' ~ container_classes %}
```
planctus, line 34: you would want to use the local variable you created here, _container_classes, but seen the previous comment maybe we could just keep this as it was...

planctus (commented):
I just left comments on the toast template, feels like those changes might be reverted, for the rest it all looks fine.

src/components/bcl-toasts/toasts.html.twig
tibi2303: done. It's reverted.

PR #455, opened 2022-09-15
feat(content-banner): Redone column classes on content-banner - OEL-2038

tibi2303:


planctus (commented):
i noticed this in the content banner with medium image, this is in the lg breakpoint:
image

tibi2303:
@planctus tried different solutions keeping in mind what you said but the problem we have it's the xxl query that does not go all the rest of size for the content area (only a portion of it). Now i simplified a bit the classes and make use of the simple 'col' class which i am using on all the md queries besides the xxl size which use the col class on lg... Hope it's clear, I tested all the different scenarios and from my point of view seems ok, please have a look yourself, thanks

PR #458, opened 2022-09-16
feat(timeline): Refactored timeline - OEL-2041

tibi2303:


src/compositions/bcl-timeline/timeline.html.twig
```diff
@@ -76,41 +74,26 @@ Parameters:
...
-            aria-expanded="false"
-            data-bs-toggle="collapse"
-            data-bs-target=".collapse-{{ _id }}"
-            class="btn btn-light btn-md"
-            type="button"
-          >
-            <span class="label-collapsed">{{ _toggle_collapsed }}</span>
-            <span class="label-expanded">{{ _toggle_expanded }}</span>
-            {% include '@oe-bcl/bcl-icon/icon.html.twig' with {
+          {% include '@oe-bcl/bcl-button/button.html.twig' with {
```
planctus, line 83: i find this unnecessary, not sure we gain anything including this component if we have to set its label like this, i feel that this was probably the reason why in the first instance this was not set as an include of the button.

tibi2303: rolled back

src/compositions/bcl-timeline/timeline.html.twig
```diff
@@ -76,24 +74,11 @@ Parameters:
...
-        {% endif %}>
-          <p class="fs-5 mb-1 fw-medium">{{ _item.label }}</p>
-          {% if _item.title is defined and _item.title is not empty %}
-          <p class="fs-6 mb-1 fw-medium">{{ _item.title }}</p>
-          {% endif %}
-          {% if _item.content is defined and _item.content is not empty %}
-          <div>{{ _item.content }}</div>
-          {% endif %}
-      </li>
+      {% include '@oe-bcl/bcl-timeline/timeline-item.html.twig' with _item|merge({
```
planctus, line 77: since we have now a dedicated template for the timeline items it feels like it should use drupal attributes and this extra_class would be defined that way.

tibi2303: done

src/compositions/bcl-timeline/timeline.html.twig
```diff
@@ -76,24 +74,11 @@ Parameters:
...
+        extra_class: extra_class,
```
planctus, line 78: i don't understand, why is this still passed as a custom param and not already as an attribute..?

planctus (approved):

PR #460, opened 2022-09-21
feat(gallery): Template splitting item and modal - OEL-2049

tibi2303:
[skip chromatic]


src/compositions/bcl-gallery/gallery-item.html.twig
```diff
@@ -0,0 +1,64 @@
...
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+{% set attributes = attributes.addClass(['bcl-gallery_item', 'd-block', 'w-100', 'img-fluid'])
+                              .setAttribute('href', '#')
+                              .setAttribute('data-bs-target', _carousel_id)
+                              .setAttribute('data-bs-slide-to', loop.index - 1)
+%}
+
+<a href="#" data-bs-target="#{{ _carousel_id }}" data-bs-slide-to="{{ loop.index - 1 }}" class="bcl-gallery__item d-block w-100 img-fluid">
```
planctus, line 37: this line is huge, can we place the attrs in a "column", instead..?

tibi2303: my mistake, here should be attributes

src/compositions/bcl-gallery/gallery-item.html.twig
```diff
@@ -0,0 +1,64 @@
...
+    {% include '@oe-bcl/bcl-icon/icon.html.twig' with {
+      name: 'play-fill',
+      size: 's',
+      path: _icon_path,
+    } only %}
+    </span>
+  {% endif %}
+  {% if _item.caption is not empty or item.caption_title is not empty %}
+    <div class="bcl-gallery__item-caption">
+    {% if _item.caption_title is not empty %}
```
planctus, line 52: i would expect some white space control, can you add a test without these info, with a caption_title but without a caption and viceversa so that you see also if twing is adding white spaces anywhere..?

src/compositions/bcl-gallery/gallery-modal.html.twig
```diff
@@ -0,0 +1,85 @@
...
+
+<div
+  {{ attributes }}
+>
+  <div class="modal-dialog modal-dialog-centered modal-fullscreen">
+    <div class="modal-content">
+      <div class="modal-header border-0 justify-content-between">
+        {% if _title is not empty %}
+          <div class="modal-title">
+            <h3 class="d-none d-sm-block">{{- _title -}}</h3>
```
planctus, line 54: i wonder if this shouldn't be configurable, how can we say that is always going to be an h3 the right tag to use..?

src/compositions/bcl-gallery/gallery-modal.html.twig
```diff
@@ -0,0 +1,85 @@
...
+          </div>
+        {% endif %}
+        <div class="carousel-pager">
+          <span>1</span> / {{ _items|length -}}
+        </div>
+        <div class="modal-close">
+          <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
```
planctus, line 61: and why this button is not an include..?

src/compositions/bcl-gallery/gallery-item.html.twig
```diff
@@ -0,0 +1,66 @@
...
+  - modal_id (string) (default: '')
+  - icon_path (string) (default: '')
+  - attributes (drupal attrs)
+#}
+
+{% set _item = item|default({}) %}
+{% set _carousel_id = carousel_id|default('') %}
+{% set _modal_id = modal_id|default('') %}
+{% set _icon_path = icon_path|default('') %}
+
```
planctus, line 24: why this empty line..?

src/compositions/bcl-gallery/gallery-item.html.twig
tibi2303: done, i updated the data to cover all the cases.

src/compositions/bcl-gallery/gallery-modal.html.twig
tibi2303: done, i included heading template.

src/compositions/bcl-gallery/gallery-item.html.twig
tibi2303: removed.

src/compositions/bcl-gallery/gallery-modal.html.twig
tibi2303: it is now, just that it feels like maybe it should have remained hardcoded.

planctus: but it's still "harcoded", it's not depending on any user data, just that if we have a template we use it instead of harcoding some html code somewhere, we do this everywhere, why do you feel it would be better otherwise here..?

src/compositions/bcl-gallery/gallery-modal.html.twig
```diff
@@ -0,0 +1,106 @@
...
+{% set _title = title|default('') %}
+{% set _title_tag = title_tag|default('h3') %}
+{% set _title_link = title_link|default({}) %}
+{% set _title_attributes = title_attributes ?: create_attribute() %}
+{% set _items = items|default([]) %}
+{% set _carousel_id = carousel_id|default('') %}
+{% set _next_label = next_label|default('Next') %}
+{% set _prev_label = prev_label|default('Previous') %}
+{% set _icon_path = icon_path|default('') %}
+
```
planctus, line 37: remove also this empty line, i mean, we don't have a strict convention about this, but as long as these are variables we set if feels this separation is not needed.

src/compositions/bcl-gallery/gallery-modal.html.twig
tibi2303: Well, this button will probably never be change. Is the x from the modal header and it could have been kept without the include in my opinion

src/compositions/bcl-gallery/gallery-modal.html.twig
```diff
@@ -0,0 +1,106 @@
...
+#}
+
+{% set _id = id|default('') %}
+{% set _title = title|default('') %}
+{% set _title_tag = title_tag|default('h3') %}
+{% set _title_link = title_link|default({}) %}
+{% set _title_attributes = title_attributes ?: create_attribute() %}
+{% set _items = items|default([]) %}
+{% set _carousel_id = carousel_id|default('') %}
+{% set _next_label = next_label|default('Next') %}
```
planctus, line 34: mmh...i don't think we have other examples of strings set as default. Generally speaking, we shouldn't, but also here this is relevant in terms of accessibility and this is a way to avoid issues, so let's keep them.

src/compositions/bcl-gallery/gallery-item.html.twig
```diff
@@ -0,0 +1,65 @@
...
+{% set _modal_id = modal_id|default('') %}
+{% set _icon_path = icon_path|default('') %}
+{% set _carousel_items = [] %}
+
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+{% set attributes = attributes.addClass(['bcl-gallery__item', 'd-block', 'w-100', 'img-fluid'])
+                              .setAttribute('href', '#')
```
planctus, line 31: why do we do this..? Is it to get the cursor changing..? I feel this could actually be a link to the image itself, in case javascript is not in place, or simply removed since the href attr is not mandatory.

tibi2303: yes, it is for the cursor, if i delete this line, it will no longer look like a pointer

tibi2303: or i can add the attribute role="button" to get this styling

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -88,32 +88,13 @@ Parameters:
               <ul class="bcl-gallery__grid">
```
planctus, line 88: i feel that instead of placing an if on the whole template, if the template is included it would at least render the wrapper, then the if would avoid rendering a lot of empty markup. Not so important, but i don't like the idea of a template that starts with a condition.

src/compositions/bcl-project/__snapshots__/project.test.js.snap
```diff
@@ -5272,13 +5272,6 @@ exports[`OE - Project page (closed) renders correctly 1`] = `
                             >
```
planctus, line 5272: there are two tests with the same "label", "renders correctly with title and link"

src/compositions/bcl-gallery/gallery-item.html.twig
```diff
@@ -0,0 +1,65 @@
...
+  - modal_id (string) (default: '')
+  - icon_path (string) (default: '')
+  - attributes (drupal attrs)
+#}
+
+{% set _item = item|default({}) %}
+{% set _carousel_id = carousel_id|default('') %}
+{% set _modal_id = modal_id|default('') %}
+{% set _icon_path = icon_path|default('') %}
+{% set _carousel_items = [] %}
```
planctus, line 24: but what is this for..? :)

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -55,10 +55,8 @@ Parameters:
```
planctus, line 55: there is _carousel_items also here, with an empty line above. The variable is not used in this template so remove the two lines all together.. :)

src/compositions/bcl-gallery/gallery-item.html.twig
tibi2303: missplaced, added from the modal template by mistake. done for both

src/compositions/bcl-gallery/gallery.html.twig
planctus: no, i didn't manage to explain myself...
The problem was having a condition at the top of a template to avoid rendering anything, this is not something we do in any other template, because it's not us that we should check that there are the bare minimum info to propoerly render a component. 
You grouped the conditions, still at the top of the template, what i expect is that these two conditions will be grouped, yes, but right before processing the items.. ;)

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -41,7 +41,6 @@ Parameters:
 {% set _items = items|default([]) %}
 {% set _icon_path = icon_path|default('') %}
 
-{% set _carousel_items = [] %}
 {% set _carousel_id = "carousel" ~ _id|capitalize %}
 {% set _modal_id = "modal" ~ _id|capitalize %}
 {% set _collpase_id = "collapse" ~ _id|capitalize %}
```
planctus, line 46: collpase_id..? typo :)

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -41,7 +41,6 @@ Parameters:
 {% set _items = items|default([]) %}
 {% set _icon_path = icon_path|default('') %}
```
planctus, line 43: would you remove this empty line as well..?

tibi2303: done, for all the points.

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -55,10 +53,7 @@ Parameters:
 
 {# Print the result #}
 
-{% if _items is not empty %}
-
   <div {{ attributes }}>
```
planctus, line 56: Can you reindent correctly the template now that we removed this initial condition..?

tibi2303: done

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -40,11 +40,9 @@ Parameters:
 {% set _max_visible_thumbnails = max_visible_thumbnails|default(5) %}
 {% set _items = items|default([]) %}
 {% set _icon_path = icon_path|default('') %}
-
-{% set _carousel_items = [] %}
 {% set _carousel_id = "carousel" ~ _id|capitalize %}
 {% set _modal_id = "modal" ~ _id|capitalize %}
-{% set _collpase_id = "collapse" ~ _id|capitalize %}
+{% set _collapsed_id = "collapse" ~ _id|capitalize %}
```
planctus, line 45: but you had to fix a typo, not rename a variable, the element is a "collapse" why is it collapse(d) and collapse(d)_id now..?

tibi2303: yes, indeed, sorry about that

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -55,153 +53,100 @@ Parameters:
...
+      } only %}
+    {% endif %}
+    </div>
+    <div class="mb-3 mb-sm-0 fw-bold">
+      {%- if _counter is not empty -%}
+        {{ _counter|replace({'%d': _items|length}) }}
+      {%- endif -%}
+    </div>
+  </div>
+  {% if _items is not empty and _items is iterable %}
```
planctus, line 74: can you do something more reindenting this template..?

tibi2303: done, i fixed also the condition on title and counter

planctus (approved):

src/compositions/bcl-gallery/gallery-item.html.twig
planctus: but we also have a css for the gallery..

tibi2303: yes, indeed, i added it there.

PR #462, opened 2022-09-27
feat(pagination): Refactored pagination - OEL-2061

tibi2303:


src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,66 @@
...
+
+{% if _path is not empty %}
+  {% set attributes = attributes.setAttribute('href', _path ) %}
+{% endif %}
+
+{% if _active %}
+  {% set attributes = attributes.setAttribute('aria-current', 'page' ) %}
+{% endif %}
+
+{% set attributes = attributes.addClass(_classes)%}
```
planctus, line 44: can you add a space before the closing tag?

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,66 @@
...
+{%- if _path is not empty -%}
+<a
+  {{ attributes }}
+>
+{%- else -%}
+<div
+  {{ attributes }}
+>
+{%- endif -%}
+  {%- if _icon is not empty -%}
```
planctus, line 55: why this indentation here..?

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,66 @@
...
+{% endif %}
+
+{% if _active %}
+  {% set attributes = attributes.setAttribute('aria-current', 'page' ) %}
+{% endif %}
+
+{% set attributes = attributes.addClass(_classes)%}
+
+{%- if _path is not empty -%}
+<a
```
planctus, line 47: can't we write it like this:
```
{%- if _path is not empty -%}
<a
{% else %}
<div
{%- endif -%}
  {{- attributes -}}
>
```

src/data/pagination/data.js
```diff
@@ -36,5 +36,16 @@ module.exports = {
       label: "3",
       path: "/example.html",
     },
+    {
+      label: "...",
```
planctus, line 40: why are we adding this..? I see this in the examples in figma, but i find it confusing because it makes you think that a pop-up would reveal the hidden items, instead here we demo it as a dead element..are we sure about doing this?

src/compositions/bcl-content-banner/content-banner.html.twig
```diff
@@ -10,8 +10,8 @@
       options: [ white, gray ]
     - image (object)
     - image_size: (string)
+      options: ['md', 'lg', 'xl']
     - date: (object)
-      options: ['sm', 'md', 'lg', 'xl']
```
planctus, line 14: but why this change is in this pull request..?

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -141,35 +134,28 @@
...
               path: _icon_path,
-            } only %}
-          {%- else -%}
-            <span aria-hidden="true">&lsaquo;</span>
-          {%- endif -%}
+            }
+          }) %}
+        {%- else -%}
+          {% set _prev = _prev|merge({
+            label: '<span aria-hidden="true">&lsaquo;</span>'
```
planctus, line 153: no no no no, we cannot do this. When do you see us storing html in a prop, tiberiu..? The problem with this is that once you will go rendering it, it will be escaped, not in storybook because we have the twig autoescape disabled, but in drupal it will. 
This can only be done for internal variables that we then render with |raw

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -104,35 +104,28 @@
...
+              icon: {
+                name: "chevron-double-left",
+                size: "fluid",
+                path: _icon_path,
+              }
+            }) %}
           {%- else -%}
-            <span aria-hidden="true">&laquo;</span>
+            {% set _first = _first|merge({
+              label: '<span aria-hidden="true">&laquo;</span>'
```
planctus, line 123: also here, i know this will make the refactoring trickier but this has to be avoided.

planctus (changes requested):
Mainly the problem is about passing html in props, there are a few other remarks, but simple ones.
I thought this would have been a deeper refactoring taking out some logic from the main template, but still we avoid lots of repeated code, that is good.

src/compositions/bcl-content-banner/content-banner.html.twig
tibi2303: This is the small update of docs regarding the content banner that we spoked on friday. Should I create a new pr only for this line?

src/data/pagination/data.js
tibi2303: We have this ticket in the backlog regarding the missing ellipsis and I thought about providing it as a demo. Also before this pr, dead elements ( with root div tags ) were not possible, we could only provide elements with anchor tag as root with href="#".

tibi2303: Maybe the behaviour could be improved but i though this should be default..

src/components/bcl-pagination/pagination-item.html.twig
tibi2303: used an internal variable for this, also for the closing tag.

src/components/bcl-pagination/pagination-item.html.twig
tibi2303: changed

src/components/bcl-pagination/pagination-item.html.twig
tibi2303: i don't no need anymore, no?

src/components/bcl-pagination/pagination.html.twig
tibi2303: Changed the approach, used the possibility for 2 different arrow (double and simple) for each left and right using a new param character_label.

src/components/bcl-pagination/pagination.html.twig
tibi2303: changed.

src/components/bcl-pagination/pagination-item.html.twig
planctus: why not tiberiu..? 
`{% set attributes = attributes.addClass(_classes) %}`

src/compositions/bcl-content-banner/content-banner.html.twig
planctus: `Should I create a new pr only for this line?` Yes, prs are for free, if you make this in a dedicated pr this fix will be tracked in a ticket and in the changelog, like this this is a custom branch and as long as we will deal with this pr the code in development will not be fixed and once we do it there will be no track of that.

src/components/bcl-pagination/pagination.html.twig
planctus: what did you do..? :) 
So now this is parametric without any default set..? You mean that in a current implementation, without setting this param, they would not have anything printed for those arrows..?
Who asked you to create the possibility of using alternative charachters, did i miss this..?
I ask you to fix something that would break and you make something that will break even more.?
It was ok like it was before you touched this, unless we received new requirements i'm not aware of.

src/data/pagination/data.js
planctus: So, basing on a ticket that never got validated, you did something in this pr that was only supposed to split the templates..?
Are we looking for troubles ourselves..? Why should we support dead elements..? Did you receive requirements about this..? If we will discuss about waht BE requested about these "dots" and we come out with clear requirements we do something about it, otherwise i don't see any reason why we shuld introduce this weak point in the component, demoing that.

tibi2303: It's removed

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,73 @@
...
+
+{% set _path = path|default('') %}
+{% set _label = label|default('') %}
+{% set _disabled = disabled|default(false) %}
+{% set _active = active|default(false) %}
+{% set _aria_label = aria_label|default('') %}
+{% set _icon = icon|default({}) %}
+{% set _position = position|default('') %}
+{% set _classes = ['page-link'] %}
+{% set _character = '' %}
```
planctus, line 24: this is a leftover

tibi2303: removed

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -9,45 +9,56 @@
...
+      - attributes (drupal attrs)
     }
-  - last (object)
+  - last (pagination Item object)
     format: {
-      - path (string)
-      - label (string)
-      - disabled (boolean)
-      - aria_label (string)
-      - icon (Icon component)
```
planctus, line 42: do we really have to repeat the same set of props for all of these objects..? can't we reduce a bit this doc comment, it seems really redundant to repeat 4 times the same data structure. Place the items one after the other and then you add a single "format" key where you detail the props

tibi2303: removed and kept only for the items, i think it should be clear enough.

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -104,35 +79,28 @@
...
+            icon: first.icon
+          }) %}
         {%- else -%}
           {%- if _icon_path is not empty -%}
-            {% include '@oe-bcl/bcl-icon/icon.html.twig' with {
-              name: "chevron-double-left",
-              size: "fluid",
-              path: _icon_path,
-            } only %}
+            {% set _first = _first|merge({
```
planctus, line 89: So..this template got simplified but now it really looks for what it is, instead of working with a single items array where to loop and include the item template, we process the four "special items" separately and in between them we process the "standard" items, now the challenge is:
can't we build a single items array in this template basing on the current data set and include the item template only once..?

tibi2303: done

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,117 @@
...
+        path: _icon_path,
+      }
+    %}
+  {% else %}
+    {% set _icon_html = '&raquo;' %}
+  {% endif %}
+{% endif %}
+
+{% if _icon is defined and _icon is not empty %}
+  {% set _label = '' %}
```
planctus, line 87: did we have this..? but in any case, why doing it..? :)

src/components/bcl-pagination/pagination.story.js
```diff
@@ -60,6 +61,17 @@ const getArgTypes = (data) => {
         category: "Style",
       },
     },
+    icon_path: {
```
planctus, line 64: mmhh....i don't understand, how is this something like a "Path to custom icons sprite", we hardcode the icon names if you pass this stuff, so this is the path to our sprite only.
It's good that you added this control and it's also good that you created it as a boolean, but the meaning of it it's not really the one you gave it. 
If you really want to be super descriptive, you can rename this in:
`path to bcl sprite file (icon_path)`
and in the description you say that
`items like prev, next, first, last will be rendered as icons instead of an html entity, it is still possible to pass a custom icon objects along those items to use custom icons.`

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,105 @@
...
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+{% if _disabled %}
+  {% set attributes = attributes.setAttribute('tabindex', '-1')
+                                .setAttribute('aria-disabled', true) 
+  %}
+{% endif %}
+
+{% if _icon is empty %}
```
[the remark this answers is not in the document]
tibi2303: indeed, it's changed

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,105 @@
...
+  {% set _icon_name = "" %}
```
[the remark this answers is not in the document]
tibi2303: added it in the first condition

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,105 @@
...
+      {% set _icon_name = 'chevron-right' %}
+      {% set _icon_html = '&rsaquo;' %}
+    {% endif %}
+  {% endif %}
+  {% set _icon = {
+    name: _icon_name,
+    size: "fluid",
+    path: _icon_path
+  } %}
+{% elseif _icon.path is empty and _icon_path is not empty %}
```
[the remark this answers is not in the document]
tibi2303: yes, changed

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,105 @@
...
+{% if _path is not empty %}
+  {% set attributes = attributes.setAttribute('href', _path ) %}
+{% endif %}
+
+{% if _active %}
+  {% set attributes = attributes.setAttribute('aria-current', 'page' ) %}
+{% endif %}
+
+{% set attributes = attributes.addClass(_classes) %}
+
```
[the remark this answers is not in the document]
tibi2303: added it in the first condition

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,103 @@
...
+      {% set _icon_name = 'chevron-right' %}
+      {% set _icon_html = '&rsaquo;' %}
+    {% endif %}
+  {% endif %}
+  {% set _icon = {
+    name: _icon_name,
+    size: "fluid",
+    path: _icon_path
+  } %}
+  {% if _icon.name is not empty %}
```
planctus, line 66: why don't you move this check, but using `_icon_name is not empty` so to include also the _icon object definition..?

tibi2303: done

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,103 @@
...
+  {% if _enable_prev_next_icon %}
+    {% if _position == 'prev' %}
+      {% set _icon_name = 'chevron-left' %}
+      {% set _icon_html = '&lsaquo;' %}
+    {% elseif _position == 'next' %}
+      {% set _icon_name = 'chevron-right' %}
+      {% set _icon_html = '&rsaquo;' %}
+    {% endif %}
+  {% endif %}
+  {% if _icon_name is not empty %}
```
planctus, line 61: here we don't know if we have a path to the icons, this is the moment when we should decide whether to build the icon object or to set the fallback html entities label.

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,103 @@
...
+  {% set attributes = attributes.setAttribute('href', _path ) %}
+{% endif %}
+
+{% if _active %}
+  {% set attributes = attributes.setAttribute('aria-current', 'page' ) %}
+{% endif %}
+
+{% set attributes = attributes.addClass(_classes) %}
+
+{% if _icon_html is not empty and _icon_path is empty %}
```
planctus, line 89: as written above this can be integrated in the main "processing" logic.
if you have the icon_path you build the object, if you don't but you have a icon_html set then you set the html entity as the label

tibi2303: done

src/components/bcl-pagination/pagination.story.js
```diff
@@ -2,14 +2,15 @@ import { withDesign } from "storybook-addon-designs";
 import withCode from "@openeuropa/storybook-addon-code";
 import { correctPaths } from "@openeuropa/bcl-story-utils";
 
-import demoData from "@openeuropa/bcl-data-pagination/data.js";
+import demoData from "@openeuropa/bcl-data-pagination/data";
```
planctus, line 5: why removing this extension..? You don't want this code to appear in the panel..?

src/components/bcl-pagination/pagination.story.js
```diff
@@ -86,7 +98,28 @@ const applyArgs = (data, args) => {
...
-  return Object.assign(data, args);
+  if (args.enable_prev_next_icon) {
+    data.prev.label = "";
+    data.next.label = "";
+  } else {
+    data.prev.label = "Previous";
+    data.next.label = "Next";
+  }
+  if (args.enable_first_last_icon) {
+    data.first.label = "";
```
planctus, line 109: is this a leftover of when we were rendering both..?

tibi2303: done, removed

src/components/bcl-pagination/pagination.story.js
tibi2303: leftover from when i tested the listing data, rolled back

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,103 @@
...
+    path: _icon_path
+  }) %}
+{% endif %}
+
+{% if _aria_label is not empty %}
+  {% set attributes = attributes.setAttribute('aria-label', _aria_label ) %}
+{% endif %}
+
+{% if _path is not empty %}
+  {% set attributes = attributes.setAttribute('href', _path ) %}
```
planctus, line 87: You can remove this, this is now happening in the link template

tibi2303: Removed, also removed the disabled, it's done also in the link

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,103 @@
...
+{% set _icon = icon|default({}) %}
+{% set _icon_path = icon_path|default('') %}
+{% set _position = position|default('') %}
+{% set _classes = ['page-link'] %}
+
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+{% if _disabled %}
```
planctus: this is also duplicated, the same happens in the link template (with an additional and useless disabled class), you can simply pass disabled to the link, then. ;)

tibi2303: done, it's removed

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -77,6 +52,30 @@
   {% set _classes = _classes ~ ' justify-content-' ~ _alignment %}
 {% endif %}
 
+{% set _internal_items = [] %}
+
+{% if _first is defined and _first is not empty %}
+  {% set _first = _first|merge({position: 'first'}) %}
```
planctus, line 58: can you add spaces after and before curly brackets..?

tibi2303: pushed

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -77,6 +52,30 @@
   {% set _classes = _classes ~ ' justify-content-' ~ _alignment %}
 {% endif %}
 
+{% set _internal_items = [] %}
```
planctus, line 55: can you move this logic under all the {% set %}, first we process all the attributes of the pagination, then we take care of the items to render

tibi2303: done

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -98,82 +97,8 @@
...
-        {%- endif -%}
-      {%- else -%}
-        {{- _prev.label -}}
-      {%- endif -%}
-      </a>
-    </li>
-  {% endif %}
-  {% if _items is not empty and _items is iterable %}
-    {% for _item in _items %}
+  {% if _internal_items is not empty and _internal_items is iterable %}
```
planctus, line 99: at line 93 we set a _class variable, it seems a leftover of something, don't even know what..

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -65,6 +41,33 @@
 {% set _last = last|default({}) %}
 {% set _items = items|default([]) %}
 {% set _icon_path = icon_path|default('') %}
+{% set _internal_items = [] %}
+
+{% if _first is defined and _first is not empty %}
```
planctus, line 46: these conditions do not make much sense, if you set the variables default then there is no need to check for `is defined`, otherwise you don't set any default and in that case you check also for the definition of the variable.

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -90,90 +89,13 @@
 {% if _aria_label is not empty %}
```
planctus, line 89: It's not clear to me the reason why we support this parameter, is it for backward compatibility..? In that case i think we should deprecate it and recommend the usage of attributes, this also means not setting it in our specs as "aria_label" but as an attribute.

tibi2303: then it's the case also for the item

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,94 @@
...
+    ['first', 'prev', 'next', 'last']
+  - attributes (drupal attrs)
+#}
+
+{% set _path = path|default('') %}
+{% set _label = label|default('') %}
+{% set _disabled = disabled|default(false) %}
+{% set _active = active|default(false) %}
+{% set _aria_label = aria_label|default('') %}
+{% set _position = position|default('') %}
```
planctus, line 24: i think we should separate internal variables from the parameters, although technically there is no difference, if you look at this doc you might think that the position parameter needs to be added and passed along the items.
Create an "internal parameters" section in the doc where to put position, enable_*, aria_label might stay again for backward compatibility but it should be deprecated in favor of attributes and in our specs this should be the way to set this attribute.

src/components/bcl-pagination/pagination.story.js
```diff
@@ -110,6 +129,5 @@ export default {
 
 export const Default = (args) =>
   pagination(applyArgs(correctPaths(demoData), args));
-
 Default.args = getArgs(demoData);
```
planctus, line 132: there is no param in this function, no need to pass the data.

tibi2303: done

src/components/bcl-pagination/pagination-item.html.twig
tibi2303: done, added also icon_path in the internal section and updated the specs for all paginations

src/components/bcl-pagination/pagination.html.twig
tibi2303: removed

src/components/bcl-pagination/pagination.html.twig
tibi2303: removed is defined from all checks

src/components/bcl-pagination/pagination.test.js
```diff
@@ -38,6 +39,32 @@ describe("OE - Pagination", () => {
...
+  });
+
+  test("without icon path renders correctly", () => {
+    expect.assertions(1);
+
+    return expect(
+      render({ ...demoData, icon_path: "", _enable_prev_next_icon: true })
+    ).resolves.toMatchSnapshot();
+  });
+
```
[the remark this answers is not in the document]
tibi2303: added a test for next with a custom icon

src/components/bcl-pagination/pagination-item.html.twig
planctus: When i said we should deprecate it, it means really adding a "deprecated, use attributes" text to its line in the doc.
But i think that you removed aria_label from the pagination template, it's the same as here, to be deprecated but not removed

tibi2303: ok, done

planctus (approved):

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,96 @@
...
+  {% endif %}
+  {% if _enable_prev_next_icon %}
+    {% if _position == 'prev' %}
+      {% set _icon_name = 'chevron-left' %}
+      {% set _icon_html = '&lsaquo;' %}
+    {% elseif _position == 'next' %}
+      {% set _icon_name = 'chevron-right' %}
+      {% set _icon_html = '&rsaquo;' %}
+    {% endif %}
+  {% endif %}
```
drishu, line 56: all these checks are overly complicating things, in the parent template, where you merge the positions, instead, just set the appropriate icon data.

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -65,22 +41,45 @@
 {% set _last = last|default({}) %}
 {% set _items = items|default([]) %}
 {% set _icon_path = icon_path|default('') %}
+{% set _internal_items = [] %}
 
-{% set _classes = 'pagination' %}
-{% if _size is not empty %}
-  {% set _classes = _classes ~ ' pagination-' ~ _size %}
+{% if _first is not empty %}
```
drishu, line 46: once you move some of the icon logic for these first last next prev items here, if some logic repeats, you could create a marco

src/components/bcl-pagination/pagination.story.js
```diff
@@ -60,6 +61,18 @@ const getArgTypes = (data) => {
         category: "Style",
       },
     },
+    icon_path: {
```
drishu, line 64: It's flouting a feature that I dont like, I'd remove this but I'll leave that up to next reviewer

src/data/pagination/data--listing.js
```diff
@@ -48,4 +49,7 @@ module.exports = {
       path: "/example.html",
     },
   ],
+  attributes: new drupalAttribute()
+    .addClass(["mt-4-5"])
```
drishu, line 53: why do we need this margin ? I know it was there before, but why

drishu (changes requested):

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,96 @@
...
+  {% if _icon_path is not empty and _icon_name is not empty %}
+    {% set _label = '' %}
+    {% set _icon = {
+      name: _icon_name,
+      size: "fluid",
+      path: _icon_path
+    } %}
+  {% else %}
+    {% if _icon_html is not empty %}
```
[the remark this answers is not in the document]
tibi2303: changed as we spoke

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,96 @@
...
+Parameters:
+  - path (string) (default: '')
+  - label (string) (default: '')
+  - disabled (boolean) (default: false)
+  - active (boolean) (default: false)
+  - aria_label (string) (default: '') - deprecated, use attributes
+  - icon (icon object) (default: {})
+  - attributes (drupal attrs)
+
+  Internal Parameters
```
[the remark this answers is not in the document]
tibi2303: changed

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,96 @@
...
+
+{% if _aria_label is not empty %}
+  {% set attributes = attributes.setAttribute('aria-label', _aria_label ) %}
+{% endif %}
+
+{% if _active %}
+  {% set attributes = attributes.setAttribute('aria-current', 'page' ) %}
+{% endif %}
+
+{% set attributes = attributes.addClass(_classes) %}
```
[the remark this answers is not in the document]
tibi2303: changed

src/components/bcl-pagination/pagination-item.html.twig
tibi2303: changed, add it in the parent and moved the icon logic in the child

src/components/bcl-pagination/pagination.html.twig
tibi2303: no need for macro anymore

src/components/bcl-pagination/pagination.story.js
tibi2303: it's just to exemplify the way icon_path logic works in the pagination, it does not impact the template in any way.

src/data/pagination/data--listing.js
tibi2303: this i am a bit uncertain to change it in the parent template.

src/components/bcl-pagination/pagination-item.html.twig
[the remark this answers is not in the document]
drishu: maybe not pushed ?

src/components/bcl-pagination/pagination.story.js
drishu: I see you insist on it, so I'll leave to a 3rd party to decide, @brummbar what do you think?

src/data/pagination/data--listing.js
drishu: ok, lets leave it for now to not impact existing sites

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,71 @@
...
+  - attributes (drupal attrs)
+
+  Internal Parameters
+  - icon_name (string) (default: '')
+  - icon_html (string) (default: '')
+#}
+
+{% set _path = path|default('') %}
+{% set _label = label|default('') %}
+{% set _disabled = disabled|default(false) %}
```
drishu, line 19: I missed these the first time around, enywhere we have booleans and we want to set defaults we should use `?? true` or `?? flase`

drishu (changes requested):

src/components/bcl-pagination/pagination-item.html.twig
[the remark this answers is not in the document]
tibi2303: If i move it when we create_attributes, the classes will not be added when the attributes comes from data

src/components/bcl-pagination/pagination-item.html.twig
tibi2303: I changed it, i will create a ticket to change it in all the templates.

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -0,0 +1,96 @@
...
+{% set _active = active|default(false) %}
+{% set _aria_label = aria_label|default('') %}
+{% set _position = position|default('') %}
+{% set _icon = icon|default({}) %}
+{% set _icon_path = icon_path|default('') %}
+{% set _enable_prev_next_icon = enable_prev_next_icon|default(false) %}
```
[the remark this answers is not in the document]
tibi2303: done, indeed, i may have missed it

src/components/bcl-pagination/pagination-item.html.twig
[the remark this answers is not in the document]
tibi2303: deleted all together

drishu (approved):

src/components/bcl-pagination/pagination.story.js
brummbar: Let's leave it for now, since it's too late to remove the whole icon idea.

PR #466, opened 2022-09-28
feat(chore): Release 0.28.0

tibi2303:


src/compositions/bcl-timeline/timeline.html.twig
```diff
@@ -17,8 +17,11 @@ Parameters:
       content: (block) (default: '')
     }
   ]
-  - hide_from (integer) (default: items.length) Item index after which to start hiding timeline items
-  - hide_to (integer) (default: items.length) Item index after which to resume displaying timeline items
+  - hide: (object)
+    format: {
+      from (integer) (default: items.length) Item index after which to start hiding timeline items
+      to (integer) (default: items.length) Item index after which to resume displaying timeline items
+    }
```
brummbar, line 24: This change is unnecessary, this is what we should avoid. Let's revert it.

tibi2303: If you are already implemented using the 'wrong' documentation, we can for sure revert this change and also modify the variables.

brummbar: Oh I didn't check that, if it's only documentation ignore indeed my comment, all good.

PR #475, opened 2022-10-11
feat(badges): Link inclusion - OEL-2066

tibi2303:


planctus:
To me, a simpler approach that would also improve in a way the markup would be to maintain the badnge as a span and including the link as an "element" of the badge. 
so `<span class="badge"><a>Link label</a><span class-"visually-hidden">Assistive text</span><span class="icon--close"></span></span>`
Doing this would not pose any risk in terms of pre-building markup chunks to be passed to the link but itwould slightlychange the behaviour, meaning that not the whole area of the badge would be clickable due to its padding.
But i think we can take care of this style so that the a tag will actually occupy the full size of the badge as it currently is leaving the chance to click on the close icon if any.

src/components/bcl-badge/badge.html.twig
```diff
@@ -22,55 +22,56 @@
...
 {% if _rounded_pill %}
   {% set _classes = _classes|merge(['rounded-pill']) %}
 {% endif %}
 
 {% if _outline %}
-  {% set _classes = _classes|merge([ 'badge-outline-' ~ _background]) %}
+  {% set _classes = _classes|merge(['badge-outline-' ~ _background]) %}
 {% else %}
-  {% set _classes = _classes|merge([ 'bg-' ~ _background]) %}
+  {% set _classes = _classes|merge(['bg-' ~ _background]) %}
```
planctus, line 34: it seems that in BS 5.2 this changed, these classes are used to set a background and color the text at once:
```
<span class="badge text-bg-primary">Primary</span>
<span class="badge text-bg-secondary">Secondary</span>
<span class="badge text-bg-success">Success</span>
<span class="badge text-bg-danger">Danger</span>
<span class="badge text-bg-warning">Warning</span>
<span class="badge text-bg-info">Info</span>
<span class="badge text-bg-light">Light</span>
<span class="badge text-bg-dark">Dark</span>
```

tibi2303: thanks, i fixed the classes, regarding the moving of the padding to the link, i would not be able to take care of focus,hover states - changing the color of the badge.

drishu (changes requested):
This is hard to review as there is a lot more here then the badge changes, and there should be nothing else

src/components/bcl-badge/badge.html.twig
```diff
@@ -22,55 +22,57 @@
...
-    {{- _label -}}
+  {{- _label -}}
   {%- if _dismissible -%}
-    <span class="icon--close" aria-hidden="true">
-      {% include '@oe-bcl/bcl-icon/icon.html.twig' with {
-        name: "x-lg",
-        path: _icon_path,
-      } only %}
-    </span>
+  <span class="{{ _icon_classes|join(' ') }}" aria-hidden="true">
```
drishu, line 56: I dont get why we need a variable for this when there is no logic modifying it before printing here

tibi2303: yes, i changed it but regarding the other changes - those are impacting also this link inclusion because of the hover effect when we have it as link so i need those changes also

tibi2303: reverted the changes and should now include only the link inclusion

drishu (approved):

src/components/bcl-badge/badge.html.twig
```diff
@@ -22,55 +22,56 @@
...
+  {% set _classes = _classes|merge(['bg-' ~ _background]) %}
 {% endif %}
 
 {% if attributes is empty %}
   {% set attributes = create_attribute() %}
 {% endif %}
 
 {%- if _url is not empty -%}
-  {% set attributes = attributes.setAttribute('title', _title).setAttribute('href', _url) %}
+  {% set attributes = attributes.setAttribute('title', _title) %}
```
brummbar, line 42: Why are we doing this? The title will be the same as the link title:
image

brummbar (changes requested):

src/components/bcl-badge/badge.html.twig
tibi2303: I changed the approach, kept the title but it's no longer the default as the label. It is only added if the title is added by the user.

brummbar (approved):

PR #479, opened 2022-10-25
feat(gallery): Raw filter used and fixed more items - OEL-2093

tibi2303:


src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -73,72 +76,79 @@ Parameters:
...
+          } only %}
+          </li>
+          {% if loop.index % 5 == 0 and not loop.last %}
+            </ul><ul class="bcl-gallery__grid">
+          {% endif %}
+        {% endfor %}
+        </ul>
+      </div>
+    {% endif %}
+    <div class="bg-lighter py-3 px-4 mt-2 rounded text-center">
```
drishu, line 115: this stuff used to be wrapped in an if so that it appears only when needed, if there ar no proper rendering tests no wonder this regression happened somewhere down the line.

src/compositions/bcl-gallery/gallery-modal.html.twig
```diff
@@ -82,8 +83,10 @@ Parameters:
         </div>
       </div>
       <div class="modal-body d-flex align-items-center justify-content-center">
+      {% block carousel %}
         {% for _item in _items %}
           {% set _carousel_items = _carousel_items|merge([_item|merge({
+            caption_title: ('<iframe' in _item.media or '<video' in _item.media) ? '' : _item.caption_title,
```
drishu, line 89: there is something wrong, not sure where the issue is coming from but:
image
if you need to repro let me know and we can do it on my Drupal instance.

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -73,72 +76,79 @@ Parameters:
...
+      <a
```
drishu, line 116: if I have more than 5 items, and I click this link, nothing happens

src/components/bcl-carousel/carousel.html.twig
```diff
@@ -120,12 +120,9 @@
     {% endif %}
     >
     {% if _item.image is defined and _item.image is not empty %}
-      {{ _item.image }}
+      {{ _item.image|raw }}
```
drishu, line 123: why do we need this ?

src/compositions/bcl-gallery/gallery-item.html.twig
```diff
@@ -51,13 +50,13 @@ Parameters:
     {% endif %}
```
drishu, line 50: I see `<p class="fs-5">{{ _item.caption_title }}</p>` but the patch has a h5, I thought first that since there is a gallery-item is the reason why we dont have the macro from the patch, but now I see this, was the patch applied or not ?

src/compositions/bcl-gallery/gallery-item.html.twig
```diff
@@ -51,13 +50,13 @@ Parameters:
...
     {% if _item.caption is not empty %}
       <div class="bcl-gallery__item-description">
-        {{ _item.caption }}
+        {{ _item.caption|raw }}
```
drishu, line 53: why do we need this raw filter everywhere ?

drishu (changes requested):
For now that's it, haven't checked if the reported issues are resolved, I'll do that after first round of feedback is implemented.

Was also thinking if it would make more sense to revert to the 24.1 variant, apply the patch, and check if there is anything else needed to fix the issues, but we need to discuss that with @brummbar

src/compositions/bcl-gallery/gallery-item.html.twig
tibi2303: It was applied, but i was thinking that we include the gallery-item and it's working in the same way as the macro, why do we still need it?

tibi2303: but for the p we change it to how it is now to simplify the heading in the library, it was some time ago between 24.1 and now - i can change it back to h5

src/compositions/bcl-gallery/gallery.html.twig
tibi2303: i can't reproduce it, when i click it the rest of the items are showing

src/compositions/bcl-gallery/gallery.html.twig
tibi2303: indeed, thanks for pointing this out, the if was only added for the container of the hidden thumbnails and not for this... i will add also a test for it

src/compositions/bcl-gallery/gallery-item.html.twig
tibi2303: I revert all of them

src/compositions/bcl-gallery/gallery-modal.html.twig
tibi2303: I tested myself on the drupal instance, I fixed it - it was because the modal had a template itself and I reverted to how it was before this template... Still weird behaviour that i could not explain is why this was happening when the template was included..

src/compositions/bcl-gallery/gallery.html.twig
drishu: works now

src/compositions/bcl-gallery/gallery-item.html.twig
drishu: not sure, I am not familiar with this change at the moment, not sure what the impact is on drupal side, we need to discuss this in the team

drishu (approved):

PR #481, opened 2022-11-03
feat(pagination): Added ellipsis support - OEL-1523

tibi2303:


src/components/bcl-pagination/pagination.html.twig
```diff
@@ -73,8 +77,27 @@
   {% endif %}
 {% endif %}
 
+
 {% set _internal_items = _internal_items|merge([_first, _prev]) %}
+
+{% if _ellipsis_before %}
+  {% set _ellipsis_before_item = {
+    icon_html: "&hellip;",
```
drishu, line 85: there is one problem with this, if icon_path is set we dont have anything:
image

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -56,12 +57,22 @@
...
+{% if _link %}
+  {%- include '@oe-bcl/bcl-link/link.html.twig' with {
+    label: _label,
+    path: _path,
+    disabled: _disabled,
+    icon: _icon,
+    attributes: attributes,
+  } only -%}
+{% else %}
+  <span
```
drishu, line 71: why not `<span {{ attributes }}>{{ _label }}</span>` do we really need all these newlines?

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -56,12 +57,22 @@
...
-{%- include '@oe-bcl/bcl-link/link.html.twig' with {
-  label: _label,
-  path: _path,
-  disabled: _disabled,
-  icon: _icon,
-  attributes: attributes.addClass(_classes),
-} only -%}
+{% set attributes = attributes.addClass(_classes) %}
+
+{% if _link %}
```
drishu, line 62: don't like all this with the link complicates the logic to much, why not just check if _path is empty or not?

drishu (changes requested):
I remembered you have some filters disabled, I need a screenshot of what it looks like without the icons when the filter is enabled in netlify.

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -56,12 +57,22 @@
...
+  {%- include '@oe-bcl/bcl-link/link.html.twig' with {
+    label: _label,
+    path: _path,
+    disabled: _disabled,
+    icon: _icon,
+    attributes: attributes,
+  } only -%}
+{% else %}
+  <span
+    {{ attributes }}
```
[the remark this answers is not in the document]
tibi2303: I need the .page-link class for the styling but indeed span inside span is not ok, i am changing the tag from span to div but keeping the attributes
Here is an img without the class added.
image

src/components/bcl-pagination/pagination-item.html.twig
tibi2303: done. also changed the span into div

src/components/bcl-pagination/pagination.html.twig
tibi2303: done.

src/components/bcl-pagination/pagination-item.html.twig
tibi2303: done. indeed

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -140,9 +140,12 @@
         {% set _item_class = _item_class|merge(['active']) %}
       {% endif %}
       <li class="{{ _item_class|join(' ') }}">
-        {% include '@oe-bcl/bcl-pagination/pagination-item.html.twig' with _item|merge({
-          icon_path: _icon_path,
-        }) only %}
+        {% if _item.icon_path is not defined %}
+          {% set _item = _item|merge({
+            icon_path: _icon_path,
```
drishu, line 145: why was this needed?

drishu (changes requested):
image
This is the reason I asked for the screenshot with the autofilter disabled, in Drupal, we dont have it disabled.

src/components/bcl-pagination/pagination.html.twig
tibi2303: if i send the icon_path to the ellipsis items without this merge, the item will search for icon_name which is not present for ellipsis. This is related to line 96 and 86 where i add for these items the icon_path as empty in order to not be added for this specific cases.

tibi2303: also i enabled the autoescaped and added the raw filter on the icon_html inside the item. This should fix that problem and this is how it looks:
image

src/components/bcl-pagination/pagination-item.html.twig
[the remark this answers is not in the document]
drishu: ok

src/components/bcl-pagination/pagination.html.twig
drishu: Something seems wrong with the enable_prev_next_icon

src/components/bcl-pagination/pagination.html.twig
drishu: I see ok

src/components/bcl-pagination/pagination-item.html.twig
```diff
@@ -33,7 +33,7 @@
 {% else %}
   {% if _icon_html is not empty %}
     {% set _label %}
-      <span aria-hidden="true">{{ _icon_html }}</span>
+      <span aria-hidden="true">{{ _icon_html|raw }}</span>
```
drishu, line 36: we controle the scope of icon_html but I still feel uneasy about this raw filter, I would have rather done something like 
```
{% set icon_html %}
      &hellip;
    {% endset %}
```
for all concerned

drishu (changes requested):

src/components/bcl-pagination/pagination-item.html.twig
tibi2303: indeed, i think it's better like you suggested and I've redone it. Still they are internal variables so it was safe before i think

drishu (changes requested):
.

drishu (approved):

PR #497, opened 2022-12-18
feat(gallery): Raw filter used and fixed more items - OEL-2093 

tibi2303:


src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -72,83 +76,134 @@ Parameters:
...
+            item: _item,
+            carousel_id: _carousel_id,
+            modal_id: _modal_id,
+            icon_path: _icon_path,
+            loop_counter: _loop_counter,
+          } only %}
+          </li>
+          {% if loop.index % 5 == 0 and not loop.last %}
+            </ul><ul class="bcl-gallery__grid">
+          {% endif %}
```
brummbar, line 111: I guess the https://twig.symfony.com/doc/3.x/filters/batch.html function is not available in twig js?

brummbar (changes requested):

src/compositions/bcl-gallery/gallery.html.twig
tibi2303: thanks for this batch, i remade the approach and it can be reviewed again, it's much better now.

PR #505, opened 2023-01-13
feat(banner): Fixed height variant on banners - OEL-2164

tibi2303:


src/compositions/bcl-banner/banner.html.twig
```diff
@@ -11,6 +11,7 @@
   - shade: (boolean) (default: false)
   - hero: (boolean) (default: false)
   - full_width: (boolean) (default: false)
+  - fixed_height: (boolean) (default: false)
```
drishu, line 14: If this is the correct behaviour for banners then I'd rather opt out than in, meaning I'd set it as true by default so everybody gets the benefit right away, and then, if its not ok, they can opt out using the parameter.
That being said, it is not specified in the requirements so I'll leave it up for discussion.

drishu (approved):

src/compositions/bcl-banner/banner.html.twig
drishu: was discussed, we stay as is

drishu (approved):

PR #509, opened 2023-01-31
feat(gallery): Improved gallery modal captions - OEL-2129

tibi2303:


src/components/bcl-carousel/carousel.html.twig
```diff
@@ -14,7 +14,8 @@
       image: (string) <img> tag
       caption (string)
       caption_classes (string)
-      caption_title (string) 
+      caption_title (string)
+      caption_visible_mobile (boolean)
```
brummbar, line 18: Value is never initialised?

src/compositions/bcl-gallery/data/data.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
       media: `<img alt="First slide" 
                 src="https://picsum.photos/id/1005/800/400" 
               />`,
+      caption_visible_mobile: true,
```
brummbar, line 20: Why should I set this when passing the information about the data?

src/compositions/bcl-gallery/gallery.html.twig
```diff
@@ -170,8 +170,6 @@ Parameters:
         {% block carousel %}
           {% for _item in _items %}
             {% set _carousel_items = _carousel_items|merge([_item|merge({
-              caption_title: ('<iframe' in _item.media or '<video' in _item.media) ? '' : _item.caption_title,
-              caption: ('<iframe' in _item.media or '<video' in _item.media) ? '' : _item.caption,
               image: ('<iframe' in _item.media or '<img' in _item.media) ? _item.media|replace({'src=': 'data-src='}) : _item.media,
```
brummbar, line 173: Side note: this breaks if there's another attribute like ``.

src/compositions/bcl-gallery/data/data.js
brummbar: This should be rather a parameter of the carousel, possibly marked as internal, which hides the caption on mobile. `show_caption_mobile` or something tied to gallery. Or better, add a block to wrap the needed code and override it when extending the carousel.

brummbar (changes requested):

src/compositions/bcl-gallery/gallery.html.twig
tibi2303: I added an extra space to be sure `src` is as alone without any extra data-. Should be ok i think

PR #515, opened 2023-03-08
feat(multiselect): Fixed overlapping border class. - OEL-2255

tibi2303:


drishu (approved):

src/themes/default/src/scss/_multiselect.scss
```diff
@@ -33,6 +33,7 @@ select.multi-select {
 
 .ss-main.multi-select {
   padding: 0;
+  border: none;
```
drishu, line 36: This didn't work for me but removing the border from ".ss-main .ss-multi-selected" did

drishu (changes requested):

src/themes/default/src/scss/_multiselect.scss
tibi2303: actually this is mostly reffering to when the .ss-main div has the form-select class which come by default on oe_bootstrap_theme.

brummbar:
We won't fix this here, but in oe_whitelabel.

Issue #517, opened 2023-03-13
Content banner card margin change


kp77:
There is a rule in [_banner.scss](https://github.com/openeuropa/bootstrap-component-library/blob/v1.0.0/src/themes/default/src/scss/_banners.scss#L141-L146) that defines horizontal margins for the `.card` inside a `.bcl-content-banner`. This rule doesn't seem to apply in the BCL examples, as the content banner doesn't have a `.container` outer element, for example see the [event page feature](https://oelibrary.netlify.app/ucpkn/?path=/story/features-event--full-page).

However, in our project, we have an outer `.container` wrapper in this case, and the rule applies. After updating to BCL 1.0.0, the previously correct image spacing in the content banner has changed, and the left edge of the image moved outside the sitewide container. See screenshot:
<img width="132" alt="Screenshot 2023-03-13 at 17 09 54" src="https://user-images.githubusercontent.com/648045/224760820-400085cc-c742-4f4a-935a-d765256d33f7.png">

I found that the `$container-padding-x` variable used to define the margins for the card was [changed in Bootstrap](https://github.com/twbs/bootstrap/commit/918a86b42528ec8f7f5d6df71851ab2ec0bc65e7) compared to the previous version we used. The value of the variable is now doubled compared to the previous Bootstrap version. Adding the following override in our theme's css (that sets the value to the previously used) fixed the alignement issue for us:

```
@include media-breakpoint-up(md) {
  .container {
    .bcl-content-banner {
      .card {
        margin-right: -$container-padding-x * .5;
        margin-left: -$container-padding-x * .5;
      }
    }
  }
}
```

kp77:
Some update on this: we had some incorrect overrides in our css due to imports that we had to fix, and that changed the situation with this issue. We still have the issue, but instead of the fix posted previously, we need to set the margins to 0 instead of `-$container-padding-x * .5`.

So for us it seems the solution would be if this rule would be removed from BCL altogether, as 0 left/right margin is the default value on the element.

kp77:
Update: the issue with the horizontal spacing is fixed after updating to BCL 1.2.0.

PR #524, opened 2023-04-13
feat(content-banner): Fix card margin inside content-banner. - OEL-2288

tibi2303:


kp77:
Tested adding the new rule on UCPKN and it fixes the issue reported in #517

PR #530, opened 2023-05-24
OEL-2772: File PDF opened in new tab.

julien-:


tibi2303:
@julien- thanks julien but i added the attributes on download button. I think this is a more general solution.
https://github.com/openeuropa/bootstrap-component-library/pull/573

Issue #533, opened 2023-07-04
Featured media wrapper issues


kp77:
The Featured media BCL composition template has a wrapper div with class `.bcl-featured-item`. I found some issues/inconsistencies (v. 1.1.0):

**1.**
The wrapper also has a `.pb-4` class in the [storybook](https://oelibrary.netlify.app/default/?path=/story/paragraphs-featured-media--featured-item&args=title_tag:h4;content_classes:bg-light+p-3), that is not present in the actual template.
**Proposed solution:**
~~The class should be removed from the storybook example to align with the template.~~
The class is [added by the wrapper_classes parameter](https://github.com/openeuropa/bootstrap-component-library/blob/v1.1.0/src/compositions/bcl-featured-media/data/data--featured-item.js#L6) for the example, but the wrapper_classes are not exposed in the Controls section for the composition, and this makes it look like as if the class would be part of the 'normal' output of the template. 

**2.**
The versions of the component that only display image/media without text (these are: [Iframe](https://oelibrary.netlify.app/default/?path=/story/paragraphs-featured-media--iframe), [Video](https://oelibrary.netlify.app/default/?path=/story/paragraphs-featured-media--video) and [Image](https://oelibrary.netlify.app/default/?path=/story/paragraphs-featured-media--image)) don't have the main component wrapper (this is because the condition in the template [only inserts](https://github.com/openeuropa/bootstrap-component-library/blob/v1.0.2/src/compositions/bcl-featured-media/featured-media.html.twig#L83-L84) the main wrapper when the `with_text` parameter is true.
**Proposed solution:**
The main wrapper should be used also when there is no text. The missing wrapper also prevents using the `wrapper_classes` parameter when there is no text (i.e. `with_text` is false).

**3.**
The component versions without text (see 2.), however, have column layout wrappers in the storybook examples, that are not in the actual template:
```
<div class="row">
  <div class="col-12 col-md-4">
    <figure class="rounded overflow-hidden d-inline-flex flex-wrap"><img class='img-fluid flex-end-100' alt='random image' src='https://picsum.photos/1000/400?random=1'>
      <figcaption class="bg-lighter p-3 flex-1 w-min-content">Media description text goes here.</figcaption>
    </figure>
  </div>
</div>
```
**Proposed solution:**
The  `<div class="row">` and  `<div class="col-12 col-md-4">` wrappers should be removed from the example to align with actual component template.

**4.**
[...]

PR #542, opened 2023-07-11
feat(featured media): Fix main wrapper when there is no text. - UCPKN-2283

kp77:
The issue was reported in point 2 of #533.


src/compositions/bcl-featured-media/featured-media.html.twig
```diff
@@ -80,8 +80,8 @@
   {% endif %}
 {% endif %}
 
-{% if _with_text %}
 <div class="{{ _classes }}">
```
tibi2303, line 83: This might have an impact on the current implementation.
The attributes are added on the figure and not on the wrapper. If someone added an utility like d-inline-block on the attributes, with what you did in this pr it will be displayed as a block, not inline.

tibi2303: we should rethink this to have no impact on the current implementation. In the end if it's only a problem on ucpkn we can always overwrite in BCL this template with your implementation

kp77: > This might have an impact on the current implementation. The attributes are added on the figure and not on the wrapper. If someone added an utility like d-inline-block on the attributes, with what you did in this pr it will be displayed as a block, not inline.

The attributes are indeed added on the figure,  but the `_classes` variable used here on the wrapper is not using the `attributes`, it is [initialized as a string](https://github.com/openeuropa/bootstrap-component-library/blob/v1.1.1/src/compositions/bcl-featured-media/featured-media.html.twig#L57) with the main class, then the `wrapper_classes` are [added to it.](https://github.com/openeuropa/bootstrap-component-library/blob/v1.1.1/src/compositions/bcl-featured-media/featured-media.html.twig#L60-L62)

So currently (without this PR) the `wrapper_classes` parameter is not working when there is no text.

kp77: Ah, sorry, probably misunderstood what you wrote. So yes, the wrapper will be added around the figure, and if the figure had custom classes, some of them should move to the wrapper_classes instead.

kp77: In case this fix is a BC problem, then I guess it has to stay like this. We don't have an issue with this in UCPKN, the reason for this PR is because I'm working on an updated version of the OE Bootstrap Theme featured media pattern, that would use the BCL component fully. The current version of the pattern is adding the main wrapper in this case.

tibi2303: ok, so what we've talked (me and francesco), he came with the solution to add a new parameter by default false ('force_wrapper') with a description - to always show the wrapper.

Keep in mind to add tests for this too - a control it may not be needed because it does not change anything in the story. thank you and what do you think? This way we can keep it backwards compatible

kp77: Hi @tibi2303,

I pushed the changes. I had to rebase anyway, because #544 was merged in the meantime. I just replaced my original commit because it was much simpler than reverting the tests and resolving the conflicts.

tibi2303 (approved):

brummbar (approved):

PR #543, opened 2023-07-11
feat(featured media): Add description title. - UCPKN-2284

kp77:
Adds support for an additional heading element that belongs to the description text of the component (the main title serves as a heading for the whole component i.e. media and text together)


src/compositions/bcl-featured-media/featured-media.story.js
```diff
@@ -54,6 +55,16 @@ const getArgTypes = (data, type) => {
         category: "Style",
       },
     },
+    description_title: {
```
tibi2303, line 58: this control does not work on my side.
image

src/compositions/bcl-featured-media/data/data--featured-item.js
```diff
@@ -4,6 +4,7 @@ const { getDummyText } = require("@openeuropa/bcl-data-utils");
 module.exports = {
   with_text: true,
   wrapper_classes: "pb-4",
+  description_title: "",
```
tibi2303, line 7: this can be deleted.

src/compositions/bcl-featured-media/featured-media.story.js
tibi2303: also it would be better to add the control in the title section.

kp77: > also we need to be aligned with the design. I will talk with them to see if we can add this.

Hi @tibi2303,

Can you let me know if there's any decision about this? (I will fix the other issues if it is  allowed, otherwise there's no point working on it.)

tibi2303: we can go forward with this 😄

tibi2303: @kp77

src/compositions/bcl-featured-media/data/data--featured-item.js
kp77: removed it.

src/compositions/bcl-featured-media/featured-media.story.js
kp77: Hi @tibi2303,

I pushed the requested changes (also rebased). 

I also changed the `title` and `tag for the title` controls to be in the 'Title' section instead of 'Content' in the 'Text with Featured Media' view. These are in the 'Title' section in the other views, and it would be confusing if they are in the Content while `description title` is in the Title.

tibi2303 (approved):

brummbar (approved):

PR #566, opened 2023-12-15
OEL-3024: Allow client to pass attributes to the pagination list.

msnassar:


drishu (approved):

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -130,7 +135,7 @@
 <nav
   {{ attributes }}
 >
-  <ul class="{{ _classes|join(' ') }}">
+  <ul {{ list_attributes }}>
```
brummbar, line 138:
```suggestion
  <ul{{ list_attributes }}>
```

src/components/bcl-pagination/pagination.html.twig
```diff
@@ -102,19 +103,23 @@
...
+  {% set list_attributes = list_attributes.addClass('pagination-' ~ _size) %}
 {% endif %}
 {% if _variant is not empty %}
-  {% set _classes = _classes|merge(['pagination--' ~ _variant]) %}
+  {% set list_attributes = list_attributes.addClass('pagination--' ~ _variant) %}
 {% endif %}
 {% if _alignment is not empty %}
-  {% set _classes = _classes|merge(['justify-content-' ~ _alignment]) %}
+  {% set list_attributes = list_attributes.addClass('justify-content-' ~ _alignment) %}
+{% endif %}
```
brummbar, line 119: In Drupal this is done this way
```
set list_classes = [
  size is not empty ? 'pagination-' ~ _size,
  _variant is not empty ? 'pagination--' ~ _variant,
  _alignment is not empty ? 'justify-content-' ~ _alignment,
]
```
Any chance we can use a similar syntax for clarity?

brummbar (changes requested):

src/components/bcl-pagination/pagination.html.twig
tibi2303: unfortunately this would create empty spaces because of the array not having default values in the params so it would look like this : ['pagination', '', '', ''] and this would create empty spaces inside the class. i wanted to create a filter like in here: https://www.drupal.org/docs/8/modules/twig-tools/sanitization-filters - remove_empty
But again twing... 

Also we have this approach in all templates.. would be a nice to have for later version for sure...

drishu (approved):

src/components/bcl-pagination/pagination.html.twig
brummbar: Ok!

PR #567, opened 2024-01-04
OEL-2784: Multiselect can not go outside of the container.

julien-:


tibi2303:
@julien- thanks julien, applied your fix plus some other improvements here: https://github.com/openeuropa/bootstrap-component-library/pull/572

PR #578, opened 2024-02-19
feat(renaming): Renaming and added deprecated badge.

tibi2303:
[skip chromatic]

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -80,7 +80,7 @@
               <a class="nav-link p-2 active" aria-current="true" href="/">Home</a>
             </li>
             <li class="nav-item col-6 col-md-auto">
-              <a class="nav-link p-2"  href="/default">BCL</a>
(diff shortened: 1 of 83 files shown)


static/index.html
```diff
@@ -80,7 +80,7 @@
               <a class="nav-link p-2 active" aria-current="true" href="/">Home</a>
             </li>
             <li class="nav-item col-6 col-md-auto">
-              <a class="nav-link p-2"  href="/default">BCL</a>
+              <a class="nav-link p-2"  href="/default">OECL</a>
```
drishu, line 83: seems a bit confusing, I think "Default" is better

drishu (changes requested):
I see the deprecated badge in the pages, can we add it also on the top, in the menu as same badge in the homepage?

static/index.html
tibi2303: okay

PR #583, opened 2024-03-21
feat(file): Download label, icon and attributes overridable OEL-2772.

julien-:


src/compositions/bcl-file/file.html.twig
```diff
@@ -138,18 +150,22 @@
...
           {% endif %}
         </div>
-        {% set _download = _download|merge({
+        {% set _download = {
           icon_position: "after",
           icon: {
             name: "download",
             size: "fluid",
             path: _icon_path,
           },
```
brummbar, line 159: Icon is not getting merged here, as it's an array so no deep merge is done. I tried locally and if I pass only name and path, size is lost. We should do a |merge(_download.icon|default({}).

src/compositions/bcl-file/file.html.twig
```diff
@@ -138,18 +150,22 @@
...
-                      .addClass(['standalone', 'align-self-center', 'd-inline-block', 'mt-1', 'mt-md-0', 'flex-shrink-0'])
-        }) %}
+          attributes: create_attribute({
+            'download' : true,
+            'target' : '_blank',
+          }|merge(_download.attributes|without('class')))
+          .addClass([
+            'standalone', 'align-self-center', 'd-inline-block', 'mt-1', 'mt-md-0', 'flex-shrink-0',
+            ]|merge(_download.attributes.class|default([]))
+          )
```
brummbar, line 167: We decided to extend the drupalAttribute class in BCL to have a matching "merge" method as in Drupal. This way merging attributes will be a breeze:
```
attributes: create_attribute({
  'download': true,
  'target': 'blank',
  'class': ['standalone', 'align-self-center', 'd-inline-block', 'mt-1', 'mt-md-0', 'flex-shrink-0'],
}.merge(_download.attributes|default(create_attribute())

PR #586, opened 2024-04-15
OEL-2918: Wrapp Gallery item with block.

julien-:


brummbar:
For this version, this has to be done in Drupal via template override.

Issue #591, opened 2024-07-01
[BCL builder] Make postcss prefixer option more flexible


vever001:
## Problem
BCL builder allows passing the `prefix` option when compiling sass files, but it is not flexible enough as it only passes the prefix and not the entire `postcss-prefix-selector` config which we would need.
Also, the plugin gets run with an empty prefix when no prefix are specified, which I believe is slowing down the compilation for no reason.

### Context:
We would need such feature/flexibility for ckeditor5 stylesheets which expects all CSS to be prefixed with ".ck-content", but we need to specify the transform callback to deal with ":root".
To be:
```
{
  entry: path.resolve(baseDir, "resources/sass/style.scss"),
  dest: path.resolve(baseDir, "assets/css/ckeditor5-style.css"),
  options: {
    ...
    prefixer: {
      prefix: ".ck-content",
      // Replace ":root" with prefix.
      transform: (prefix, selector, prefixedSelector) => {
        if (selector.includes(":root")) {
          return selector.replace(/:root/g, prefix);
        }
        return prefixedSelector;
      }
    }
  }
}
```

## Proposed resolution
Deprecate the `prefix` option in favor of `prefixer` which holds the entire prefixer config.
Run the plugin only when one of these options is passed.

tibi2303:
ok, i am going to recreate the same pr on bcl directly, the solution seems good on my side

vever001:
Why? Sorry but this isn't promoting contributions... no cherry picking...
I understand you may need to further refine the code or create a dedicated branch but in this case the code is exactly the same between PRs...

tibi2303:
Recreated because of checks - in order to have the tests pass.. sorry for this. I saw you got 500 internal server error..
https://github.com/openeuropa/bootstrap-component-library/pull/610

PR #592, opened 2024-07-01
feat: Make postcss prefixer option more flexible.

vever001:
See https://github.com/openeuropa/bootstrap-component-library/issues/591


vever001:
All tests should pass but it got a: Request failed "500 Internal Server Error"

tibi2303:
ok, i am going to recreate the same pr on bcl directly, the solution seems good on my side

PR #593, opened 2024-07-03
OEL-3093: Card - Text with breakline.

julien-:


brummbar:
A new line is not a line break. This is not to be done in the component, it will horribly break a lot of websites.

PR #618, opened 2024-10-23
OEL-3407: Added focus state on accordion.

tibi2303:
[skip chromatic]

### DIFF
--- a/src/themes/default/src/scss/_accordion.scss
+++ b/src/themes/default/src/scss/_accordion.scss
@@ -1,5 +1,5 @@
 .accordion-button {
-  &:focus {
+  &:focus-visible {
     box-shadow: none;
     border: none;
     outline: 2px solid $primary;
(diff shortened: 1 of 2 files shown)


src/themes/default/src/scss/_accordion.scss
```diff
@@ -0,0 +1,7 @@
+.accordion-button {
```
drishu, line 1: as far as I can see the button:focus:not(:focus-visible) from reboot css is preventing this style from being displayed

drishu (changes requested):

src/themes/default/src/scss/_accordion.scss
tibi2303: yes, indeed.. after some investigation we should use :focus-visible.

drishu (approved):

PR #624, opened 2024-10-29
OEL-3401: Input focus state.

tibi2303:
[skip chromatic]

### DIFF
--- a/src/themes/default/src/scss/_input.scss
+++ b/src/themes/default/src/scss/_input.scss
@@ -1,11 +1,16 @@
-// stylelint-disable no-duplicate-selectors
-// stylelint-disable selector-no-qualifying-type
+/* stylelint-disable no-duplicate-selectors, selector-no-qualifying-type, no-descending-specificity */
 
 .form-select {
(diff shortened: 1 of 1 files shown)


src/themes/default/src/scss/_input.scss
```diff
@@ -30,6 +41,20 @@
     margin-top: 0.15rem;
   }
 }
+.form-check-input {
+  &:focus-visible {
+    outline: 4px solid $primary;
```
drishu, line 46: why is this 4px, might be in figma but dont have access, can you provide screenshot to be clear

drishu (changes requested):

src/themes/default/src/scss/_input.scss
tibi2303: image
image

yes, it should be 4px

tibi2303: also here are the designs for other form elements:
image
image
image

drishu (approved):

PR #629, opened 2024-11-27
OEL-3552: Check condition for accessible toggle.

tibi2303:


drishu (changes requested):
This branch as far as I can see was done from development, it should be rebased to the latest tag instead

PR #632, opened 2024-12-04
OEL-3547: Added collapse/expand buttons in accordion.

tibi2303:
[skip chromatic]


src/components/bcl-accordion/accordion.html.twig
```diff
@@ -43,57 +47,89 @@
...
+          class="accordion-header" 
+          id="heading-{{ _id }}-{{ loop.index }}"
+        >
+          {% set button_attributes = create_attribute()
+                                    .addClass(_button_classes)
+                                    .setAttribute('data-bs-toggle', 'collapse')
+                                    .setAttribute('autocomplete', 'off')
+                                    .setAttribute('data-bs-target', '#collapse-' ~ _id ~ '-' ~ loop.index)
+                                    .setAttribute('aria-controls', 'collapse-' ~ _id ~ '-' ~ loop.index)
+                                    .setAttribute('aria-expanded', open_item ? 'true' : 'false')
```
drishu, line 108: think we where doing all these in one go 
```
create_attribute([
..
])
```
everywhere needed pls

drishu (changes requested):

src/components/bcl-accordion/accordion.html.twig
tibi2303: yes, indeed it would be better, but the plugin we use ('twing') for the compiling in js of twig file does not support this format of create_attribute... i know we should use the approach you suggested but in all the twig file we use it like we have it now (with .setAttribute).. maybe a different ticket to investigate and change everywhere to the approach you suggested would be better

src/components/bcl-accordion/accordion.html.twig
```diff
@@ -43,57 +47,89 @@
...
+                    .setAttribute('data-action', 'collapse')
+      }) only %}
     {% endif %}
-    >
-      <div class="accordion-body">
-        {%- set _content = _item.content|default('') %}
-        {%- block content _content -%}
+  </div>
+  {% endif %}
+  <div class="accordion-items-wrapper">
```
drishu, line 89: I tested and we can drop this

src/components/bcl-accordion/accordion.html.twig
```diff
@@ -43,57 +47,89 @@
...
-    </{{ _item_title_tag }}>
-    <div
-      id="collapse-{{ _id }}-{{ loop.index }}"
-      class="accordion-collapse collapse{{ _open_item ? ' show' }}"
-      aria-labelledby="heading-{{ _id }}-{{ loop.index }}"
-      role="region"
-    {% if not _item.stay_open %}
-      data-bs-parent="#accordion-{{ _id }}"
+  {% if _items|length > 1 and (_expand_button is not empty or _collapse_button is not empty) %}
+  <div class="d-flex justify-content-end gap-3 mb-3">
```
drishu, line 60: Can we transfer these to a class?

src/components/bcl-accordion/accordion.html.twig
```diff
@@ -43,57 +47,89 @@
...
+            clean_class: true,
+            attributes: button_attributes
+          } only %}
+        </{{ _item_title_tag }}>
+        <div
+          id="collapse-{{ _id }}-{{ loop.index }}"
+          class="accordion-collapse collapse{{ _open_item ? ' show' }}"
+          aria-labelledby="heading-{{ _id }}-{{ loop.index }}"
+          role="region"
+        {% if not _item.stay_open and _expand_button is empty  %}
```
drishu, line 121: `and _expand_button is empty` why is this needed?

drishu (changes requested):

src/components/bcl-accordion/accordion.html.twig
tibi2303: because this attribute helps the accordion to close other 'tab' when another is opened. And when expand_button is clicked, the behaviour of the accordion does not know how to open all of them, only one because of this attribute. For the collapse_button i dont care, because it only closses all.

tibi2303: when this attribute is not present, all the tabs can be opened separately

src/components/bcl-accordion/accordion.html.twig
tibi2303: i created an attribute for them.

src/components/bcl-accordion/accordion.html.twig
tibi2303: deleted.

tibi2303: this is for the default bootstrap styles
image

on :first-of-type on accordion items it adds a border, when we have the buttons, it no longer adds the border styles because it is not :first-of-type, so i encaspsulate them in a div to have the styles for border

piotrsmykaj (approved):

drishu (commented):
2 quick on that merge trigger

drishu (approved):

PR #643, opened 2025-02-14
ERATRANS-1057: Mega menu.

tibi2303:


tibi2303:
Prerelease done: 0.1057.202502162340.

src/compositions/bcl-mega-menu/mega-menu.html.twig
```diff
@@ -0,0 +1,112 @@
...
+  - icon_path (string) (default: '')
+#}
+
+{% set _id = id|default('dropdown-' ~ random(1000)) %}
+{% set _link = link ?? false %}
+{% set _trigger = trigger|default({}) %}
+{% set _content_block = content_block|default('') %}
+{% set _items = items|default([]) %}
+{% set _back_button_label = back_button_label|default('Back') %}
+{% set _icon_path = icon_path|default('') %}
```
drishu, line 26: add a min height parameter with a default value to what is now in ecl. If the content height is bigger, then scrolling is fine

src/compositions/bcl-mega-menu/mega-menu.html.twig
```diff
@@ -0,0 +1,112 @@
...
+        <div class="content-block py-3">
+          {%- block content_block -%}
+            {{ _content_block }}
+          {%- endblock -%}
+        </div>
+      </div>
+      {% endif %}
+      <div class="col-12 col-lg-8">
+        <div class="navigation-items">
+          {% for _item in _items %}
```
drishu, line 87: look into a solution, css prefereply with some pseudo code or js if not, where if the first item in any column inside the meganenu has children, then we open it automatically 1st and 2nd level both (see ecl example)

src/compositions/bcl-mega-menu/mega-menu.html.twig
```diff
@@ -0,0 +1,112 @@
...
+            {% if _item.items is defined %}
+              {% set _trigger_attributes = _item.trigger.attributes is defined ? _item.trigger.attributes : create_attribute() %}
+              {% include '@oe-bcl/bcl-dropdown/dropdown.html.twig' with _item|merge({
```
drishu, line 90: make sure that columns can be targeted separately in css if needed also toghter with their parent link

src/components/bcl-navigation/navigation.html.twig
```diff
@@ -124,6 +124,8 @@
           {% include '@oe-bcl/bcl-dropdown/dropdown.html.twig' with _item only %}
         {% elseif _item.button is defined %}
           {% include '@oe-bcl/bcl-button/button.html.twig' with _item only %}
+        {% elseif _item.mega_menu is defined %}
+          {% include '@oe-bcl/bcl-mega-menu/mega-menu.html.twig' with _item only %}
```
drishu, line 128: double check that it renders correclty when used inside a menu

src/compositions/bcl-mega-menu/mega-menu.html.twig
```diff
@@ -0,0 +1,112 @@
...
+
+{% if attributes is empty %}
+  {% set attributes = create_attribute() %}
+{% endif %}
+
+{% set _classes = ['bcl-mega-menu'] %}
+
+{% set attributes = attributes.setAttribute('aria-labelledby', _id).addClass(_classes) %}
+
+{% if _trigger.attributes is empty %}
```
drishu, line 36: when on tabled mode, the parent data should be displayed on top
image

src/compositions/bcl-mega-menu/mega-menu.html.twig
```diff
@@ -0,0 +1,112 @@
...
+              {% include '@oe-bcl/bcl-dropdown/dropdown.html.twig' with _item|merge({
+                direction: 'dropend',
+                link: true,
+                trigger: _item.trigger|merge({
+                  attributes: _trigger_attributes.setAttribute('data-bs-auto-close', 'outside')
+                })
+              }) only %}
+              {% if _item.collapse is defined %}
+                <div class="collapse">
+                  {{ _item.collapse|raw }}
```
drishu, line 99: what is this collapse|raw? please give me an example

src/themes/default/src/scss/_mega-menu.scss
```diff
@@ -0,0 +1,166 @@
+/* stylelint-disable no-descending-specificity, declaration-no-important */
```
drishu, line 1: the second column should be white and the third gray as in ecl

drishu (changes requested):

src/compositions/bcl-mega-menu/mega-menu.html.twig
tibi2303: It is used to render properly the html... the encoding if the |raw is not added sees the collapse parameter as html and does not render it.. that's why i used raw here

src/compositions/bcl-mega-menu/mega-menu.html.twig
tibi2303: added. it's pixel based, the user can add any size but will default to 300px

src/compositions/bcl-mega-menu/mega-menu.html.twig
tibi2303: they can, we have .content-block for first column, .navigation-items for second column, .navigation-items .dropdown-menu for third column and .collapse for 3rd column

src/components/bcl-navigation/navigation.html.twig
tibi2303: yes, fixed and now renders okay inside a menu

src/compositions/bcl-mega-menu/mega-menu.html.twig
tibi2303: added

src/themes/default/src/scss/_mega-menu.scss
tibi2303: changed the colors

PR #644, opened 2025-02-20
feat(storybook): Update dependencies and storybook to vers 8. - OEL-3643

tibi2303:
[skip actions]


drishu (approved):
There still are messages about dependencies (their versions) but if the scope is to remove deprecations then it seems to be done

PR #661, opened 2025-04-11
refactor: OEL-3775: Negative extended spacers.

julien-:
At this moment it exists negative bootstrap classes :

`.mt-n4` for a margin top negative of value 4 (which is "2 rem" in CSS value)
It exists` .mt-4-75` but not `.mt-n4-75`


tibi2303:
Thanks for this, i will recreate just to have the deploy and checks pass

tibi2303:
https://github.com/openeuropa/bootstrap-component-library/pull/662

PR #665, opened 2025-04-16
feat(accordion): Limit to 80ch per row. - OEL-3656

tibi2303:


src/themes/default/src/scss/_accordion.scss
```diff
@@ -5,3 +5,6 @@
     outline: 2px solid $primary;
   }
 }
+.accordion-body {
+  max-width: 80ch;
```
AaronGilMartinez, line 9: This question is more curiosity than a review question since the technical analysis is already done. Why _ch_ and not the bootstrap grid system  since you want to restrict to 9 columns? This characters thing varies depending on fonts as you comment in the ticket.

AaronGilMartinez (changes requested):

src/themes/default/src/scss/_accordion.scss
tibi2303: it is proposed to use the same approach as ecl.. they have 80ch limit
https://ec.europa.eu/component-library/playground/ec/?path=/story/components-accordion--default&globals=viewport:responsive

src/themes/default/src/scss/_accordion.scss
```diff
@@ -5,3 +5,6 @@
     outline: 2px solid $primary;
   }
 }
+.accordion-body {
```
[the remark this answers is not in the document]
tibi2303: yes, thanks indeed.

AaronGilMartinez (approved):

PR #666, opened 2025-04-16
feat(ids): Lower probability of same id twice on page. - OEL-3642

tibi2303:


drishu (changes requested):
We are missing the in-page navigation

drishu (approved):

PR #671, opened 2025-04-28
OEL-3679: Changed sidebar heading tag to button.

tibi2303:


src/themes/default/src/scss/base/_layout.scss
```diff
@@ -1,4 +1,5 @@
 @include media-breakpoint-up(lg) {
```
AaronGilMartinez, line 1: Is this font size jump ok? The cols change to sidebar is using md.

AaronGilMartinez (changes requested):

src/themes/default/src/scss/base/_layout.scss
vengador: Fixed SCSS rules.

AaronGilMartinez (approved):

PR #672, opened 2025-04-28
OEL-3694: Mega menu.

tibi2303:


src/data/navbar/data.js
```diff
@@ -10,6 +10,107 @@ module.exports = {
...
     items: [
+      {
+        mega_menu: true,
+        link: true,
+        icon_path: "/icons.svg",
+        id: "dropdown-megamenu",
+        trigger: {
+          label: "A link with dropdown",
+          path: "/example.html",
+          attributes: new drupalAttribute().addClass("nav-link"),
```
donquixote, line 21: Let's add such classes automatically, not have them passed in.

src/compositions/bcl-mega-menu/mega-menu.html.twig
```diff
@@ -0,0 +1,215 @@
...
+                  {% if _raw.items is defined and _raw.items is not empty %}
+                    {% set _panel_id = 'mm-panel-' ~ _id ~ '-' ~ loop.index0 %}
+                    {% set _trigger_id = 'mm-trigger-' ~ _id ~ '-' ~ loop.index0 %}
+
+                    {% set _raw_items = _raw.items %}
+
+                    {% if _raw.see_all is defined and _raw.see_all is not empty %}
+                        {% set _see_all_attr = (_raw.see_all.attributes is defined ? _raw.see_all.attributes : create_attribute()) %}
+                        {% set _see_all = _raw.see_all|merge({
+                          label: '<span>' ~ _raw.see_all.label ~ '</span>',
```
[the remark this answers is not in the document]
donquixote: Also why do we need a `<span>` here?
I would say we only need a span when there is no link tag (`<a>`), so for nolink items. The "See all" is always a link.

src/themes/default/src/scss/_header.scss
```diff
@@ -195,18 +112,47 @@ $header-link-active-background: #003776 !default;
       position: static;
     }
   }
+  .bcl-header__top-navbar {
```
donquixote, line 115: What are we targeting here? I don't see this class in the Drupal page.

src/themes/default/src/scss/_header.scss
```diff
@@ -195,18 +112,47 @@ $header-link-active-background: #003776 !default;
...
-      max-width: 100%;
-      .nav-item {
-        &.dropdown {
-          max-width: 100%;
-          a {
-            white-space: normal;
-          }
-          .dropdown-menu {
-            max-width: 100%;
+  .bcl-header__navbar.navbar {
```
donquixote, line 119: Also this here, I don't see it in the Drupal page.

src/themes/default/src/scss/_header.scss
```diff
@@ -252,12 +219,72 @@ $header-link-active-background: #003776 !default;
...
+        width: 10px;
+        height: 10px;
+        border: none;
+        margin-top: 4px;
+      }
+      &:hover,
+      &:active,
+      &:focus {
+        &:after {
+          background-color: $primary;
```
donquixote, line 271: With one item being "active", another being "open", and a third in "focus", we get 3 items with white background. This can look quite confusing.

donquixote:
Some more general feedback (2025-09-05):

Mobile / narrow viewport appearance:
- The margins left/right in mobile seem too big - compare with ECL and Figma, talk with Teodora if needed.
  I think we want the menu to align with other content elements.
  (Also review these margins in medium and large viewport width)
- The Figma design has a color gradient in the blue background area in mobile.
- In newest Figma and in ECL, every submenu layer completely covers what is in the background. The regular page content is completely hidden when the menu is opened, and each submenu level fully covers the parent (sub)menu level.

General appearance:
- I think the idea is that links should have blue text even when they are not hovered, and even in the first submenu level. On the other hand, non-link parent items should never have blue text or underline, no matter if open or hovered. A goal here, besides following the design, is to make links distinguishable from non-link buttons.
- The first submenu level has no minimum height and shrinks if it has only few items. The second level submenu has a minimum height. Once it is opened, the mega menu no longer shrinks.

Accessibility:
The following points should be checked with Teodora.
- Currently the first submenu level does not use ul/li.
- The second level submenus are not direct children of their parent item. We use aria attributes and tabindex to mitigate this, which actually seems to work surprisingly ok.
- The second level submenu can be opened by clicking the parent item. But clicking it again does not close it. This is different from how these menus usually work. (And due to us using tabs, technically)

src/themes/default/src/scss/_mega-menu.scss
```diff
@@ -0,0 +1,423 @@
...
+              width: 100%;
+            }
+            svg {
+              width: 0.8rem;
+            }
+            span {
+              text-overflow: ellipsis;
+              overflow: hidden;
+              display: block;
+              max-width: 85%;
```
donquixote, line 186: You can drop this max-width if you set `flex-shrink: 0;` on the svg.
Then you can use `gap` on the parent (link) element, instead of the margin-left on the svg.

src/themes/default/src/scss/_mega-menu.scss
```diff
@@ -0,0 +1,423 @@
...
+          width: 100%;
+          height: 100%;
+          content: "";
+          z-index: 1060;
+        }
+      }
+      header {
+        position: relative;
+        z-index: 1070;
+      }
```
donquixote, line 336: This is too generic, it selects every header element on the page.\
It will be a problem for themes that are not based on oe_whitelabel, or that have customizations.\
(E.g. in oe_bootstrap_theme there is a header element in the top of main)

One solution could be to further qualify the selector e.g. with another `header:has(.mega-menu-container.show)`.

But I would prefer something more local and modular, that does not depend on a specific page structure.\
Thinking about it...

donquixote:
Something about scrolling.
Currently we are disabling the page scroll when the menu is revealed.
The page is locked in the current scroll position, which can be undesirable in some scenarios:
- If the navigation bar is half visible at the top end of the viewport, it will remain so when the megamenu is opened.
- If the page contains a huge header element (e.g. an image) above the navigation bar, then the navigation bar could be at the bottom of the viewport, and you cannot scroll down to reveal all of the mega menu.
- If the viewport height is very short (unlikely but possible), the mega menu won't be fully visible and you cannot scroll down to fully reveal it. 

How do other mega menu implementations handle this?
- ECL jumps to the top of the page when you open a mega menu item.
  I personally don't like this jump.
- In some other mega menus, the navigation bar is has position fixed or sticky, so it remains at the top of the viewport.
  This does not fix the last problem with short viewport height, but maybe it is rare enough that we can accept it.

src/themes/default/src/scss/_mega-menu.scss
donquixote: For he overlay and z-index problem, it depends on whether we want a page-level vertical scrollbar.

### Goal
The goal is to cover everything below the navigation bar (main page content and footer), but not the navigation bar itself or anything above it.

### Solution without page-level scrollbar
Goal:
- Cover elements as above.
- No page-level vertical scrollbar when mega menu is opened (as currently the case).

The solution is simple:
Attach the overlay to the mega menu itself, e.g. as `.mega-menu-container.dropdown-menu::after`.
Give it position: absolute, height: 100vw or even more, and let it begin at the top or bottom of the megamenu.
The overlay gets the same z-index as the megamenu itself, which is good enough to cover anything further down in the page.

### Solution with page-level scrollbar
Goal:
- Cover elements as above.
- The page-level vertical scrollbar remains enabled when megamenu is opened.
- _The overlay should not impact the available scroll length. (without this requirement, it is easy)_

Solution:
We create an overlay that covers the entire page (position: absolute), or the entire viewport (position: fixed).
The overlay could be attached to the `body` element or anything else.
Now we need to set a z-index on the header or other element that contains the megamenu, and all elements before it.
It is easy in oe_whitelabel, where we know the page structure.
In oe_bootstrap_theme there is no `<header>` element in that place, only a `<nav>`.
So it is hard to select these elements.

tibi2303: thanks andreas, that's what i already did... check selector on line 332

src/themes/default/src/scss/_mega-menu.scss
tibi2303: thanks, it's a viable fix

tibi2303:
The other points you can also check, they are fixed. @donquixote

src/themes/default/src/scss/_mega-menu.scss
donquixote: You are now using flex-shrink 0, but still use margin instead of gap..
(In this case the effect is the same, but gap is generally more flexible and easier to apply in all situations.)

src/compositions/bcl-mega-menu/mega-menu.html.twig
```diff
@@ -0,0 +1,218 @@
...
+        </div>
+        {% endif %}
+        <div class="col-md-12 col-xl-6">
+          <div class="row g-0">
+            <div class="col-12 col-lg-4 col-xl mt-lg-3 mt-xl-0">
+              <div class="navigation-items" role="tablist" aria-orientation="vertical">
+                {% set _dropdown_panels = _dropdown_panels|default([]) %}
+
+                {% for _raw in _items %}
+                  {% set _render_item = _raw %}
```
donquixote, line 116: This variable is confusing.
It can be more clear if we fully split up the two cases, and have separate include statements:

Version 1, here we still use a link for the parent item:
```twig
{% if _item.items is defined and _item.items is not empty %}
  # This is a parent item with a trigger to open the submenu.
  [..]
  {% set _trigger_link = ... &}
  {% include '..link.html.twig' with _trigger_link only %}
{% else %}
  # This is a leaf item.
  {% include '..link.html.twig' with _item only %}
{% endif %}
```

Version 2, with:
- Condition can be for '_item.trigger` instead of `_item.items`.
- We print a button instead of a link for the trigger.

```twig
{% if _item.trigger is defined and _item.trigger is not empty %}
  # This is a parent item with a trigger to open the submenu.
  [..]
  {% set _trigger_button = ... &}
  {% include '..button.html.twig' with _trigger_button only %}
{% else %}
  # This is a leaf item.
  {% include '..link.html.twig' with _item only %}
{% endif %}
```

This is just for a general idea.

src/themes/default/src/scss/_header.scss
```diff
@@ -4,84 +4,23 @@ $header-background-start: #039 !default;
...
-      &:focus,
-      &:hover,
-      &.active {
-        color: #fff;
-      }
-      &:focus-visible {
-        outline: 2px solid $white;
-        border-radius: 4px;
-        outline-offset: 0;
-      }
```
donquixote, line 60: This style used to apply on all viewports, now we change it so it only applies in desktop.
In mobile we don't get a proper outline anymore.

src/themes/default/src/js/mega-menu/mega-menu.js
```diff
@@ -0,0 +1,174 @@
...
+   * Closes all submenus.
+   *
+   * This is simple while there is only one submenu level.
+   *
+   * @returns {HTMLElement[]}
+   *   Triggers for submenus that were closed.
+   *   Usually this is either exactly one, or none.
+   */
+  closeAllSubmenus() {
+    const triggers = SelectorEngine
```
piotrsmykaj, line 151:
```suggestion
    if (!this.ul) {
      return;
    }
    const triggers = SelectorEngine
```

piotrsmykaj (approved):

Issue #657, opened 2025-04-08
Pattern pagination won't wrap


Aporie:
A small improvement to be added to the pattern pagination.

When having a lot of elements and on mobile, the navigation bar may overflow the content.

![Image](https://github.com/user-attachments/assets/31b05996-34c8-4ea2-b153-3c647748a077)

Aporie:
[pattern_pagination.txt](https://github.com/user-attachments/files/19645561/pattern_pagination.txt)

I guess the "display: flex;" is coming from @openeuropa/bcl-bootstrap/scss/pagination, but this could be repeated here to make sure we have a flex on the pagination.

tibi2303:
Sorry for late response, we will move to pagination v2 in this pr:
https://github.com/openeuropa/bootstrap-component-library/pull/701

and pagination v1 will be deprecated and removed when all the projects will manage to move to v2. Can you please check and let me know that v2 comply with flex display?

PR #674, opened 2025-04-29
ELA-1306: Wrap pager for navigation.

Aporie:
See issue #657


tibi2303:
this can be a quick solution, though it does not look okay on very small viewports.. we need to speak to the designers to adapt this view.. i will approve.

tibi2303 (approved):

PR #684, opened 2025-07-14
OEL-2127: Add visually-hidden element when there is no label for close button in offcanvas

vengador:


src/components/bcl-offcanvas/offcanvas.html.twig
```diff
@@ -89,18 +90,28 @@
...
     {% if _extra_classes_close is not empty %}
-      {% set button_attributes = button_attributes.addClass(_extra_classes_close) %}
+      {% set _close_attributes = _close_attributes.addClass(_extra_classes_close) %}
+    {% endif %}
+
+    {% if _close_aria_label %}
+      {% set _close_attributes = _close_attributes.setAttribute('aria-label', _close_aria_label) %}
+    {% else %}
+      {% set _close_content %}
+        <span class="label visually-hidden">{{ "Close" }}</span>
```
AaronGilMartinez, line 107: I wonder if we can add the translation filter here and if we should, this texts may not get translated.

AaronGilMartinez (changes requested):

PR #702, opened 2025-10-30
OEL-4185: Improved mega menu.

tibi2303:


src/components/bcl-navbar/__snapshots__/navbar.test.js.snap
```diff
@@ -148,6 +148,7 @@ exports[`OE - navbar dark renders correctly 1`] = `
                             <div
                               aria-labelledby="custom-trigger-mm-1"
                               class="bcl-mega-menu__submenu __level-2"
+                              hidden="true"
```
donquixote: This looks wrong. The `hidden` attribute should not have a value, at least not in its default usage.
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden
(see below, I will also comment in the respective template)

src/themes/default/src/scss/_mega-menu.scss
```diff
@@ -190,6 +190,29 @@ ul.bcl-mega-menu__items {
...
+    // Use flex for icon spacing.
+    display: flex;
+    align-items: center;
+    color: $link-color;
+    @include media-breakpoint-up(lg) {
+      padding: $mm-gutter-y map-get($spacers, "4-25");
+      margin: 0 -1.75rem;
+    }
+    &:hover {
+      background: $primary-bg-subtle;
```
donquixote: This does nothing. The background is already that exact color.
It would make a difference if it was in the parent menu.

src/themes/default/src/scss/_mega-menu.scss
```diff
@@ -190,6 +190,29 @@ ul.bcl-mega-menu__items {
   margin: 0;
   padding-inline-start: 0;
   // Add a scrollbar in desktop viewport.
+  .see-all-button {
+    @include mm-item-padding();
+    // Use flex for icon spacing.
+    display: flex;
```
donquixote: Note:
This used to have `justify-content: space-between`, which pushed the icon to the right.
By dropping that rule, we align closer with Figma.
-> good :)

src/themes/default/src/scss/_mega-menu.scss
donquixote: Actually it does make a difference in mobile.
-> let's keep it :)

src/themes/default/src/scss/_mega-menu.scss
```diff
@@ -190,6 +190,29 @@ ul.bcl-mega-menu__items {
   margin: 0;
   padding-inline-start: 0;
   // Add a scrollbar in desktop viewport.
```
donquixote: This comment is now removed from its target.

donquixote (commented):
@tibi2303 I left some comments but also pushed some commits.
Let me know if you agree with these.

src/themes/default/src/js/header/header.js
```diff
@@ -0,0 +1,98 @@
...
+
+    this.init();
+  }
+
+  init() {
+    this.updateOffset();
+
+    EventHandler.on(window, "resize", this.boundScheduleUpdate);
+    EventHandler.on(window, "orientationchange", this.boundScheduleUpdate);
+    EventHandler.on(window, "scroll", this.boundScheduleUpdate);
```
donquixote: Do we need the scroll trigger? Maybe it is for some websites where a top banner changes size on scroll?

src/themes/default/src/js/header/header.js
```diff
@@ -0,0 +1,98 @@
...
+
+    const togglers = SelectorEngine.find(TOGGLER_SELECTOR, this.element);
+    togglers.forEach((toggler) => {
+      EventHandler.on(toggler, "click", this.boundScheduleUpdate);
+    });
```
donquixote: I think in default layout clicking that button does not change the height. But ok if we want to support other scenarios.

src/themes/default/src/js/header/header.js
```diff
@@ -0,0 +1,98 @@
...
+
+  scheduleUpdate() {
+    if (this.frameRequest) {
+      return;
+    }
+
+    this.frameRequest = window.requestAnimationFrame(() => {
+      this.frameRequest = null;
+      this.updateOffset();
+    });
```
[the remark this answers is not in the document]
tibi2303: yes, only when finished with resize it will recalculate

src/themes/default/src/js/header/header.js
```diff
@@ -0,0 +1,98 @@
...
+
+    this.boundScheduleUpdate = this.scheduleUpdate.bind(this);
+
+    this.init();
+  }
+
+  init() {
+    this.updateOffset();
+
+    EventHandler.on(window, "resize", this.boundScheduleUpdate);
```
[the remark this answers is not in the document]
tibi2303: indeed, added directly scheduleUpdate

donquixote (approved):
Looks good!
But let'd do UAT release before we merge.

PR #704, opened 2025-11-03
OEL-4278: Extending drupal attributes and fixes for proper html attributes.

tibi2303:


src/components/bcl-card-layout/__snapshots__/card-layout.test.js.snap
```diff
@@ -585,7 +585,7 @@ exports[`OE - Card grid renders correctly with cards having equal height 1`] = `
       class="col"
     >
       <article
-        class="mb-3 card h-100"
+        class="mb-3 h-100 card"
```
donquixote: What causes the classes to get reordered?

tools/test-utils/index.js
```diff
@@ -1,22 +1,192 @@
+const path = require("path");
+const prettyFormat = require("pretty-format");
```
donquixote: My IDE tells me that `pretty-format` package should be explicitly required in package.json.
I assume that before now we did not directly use it, and we get it as an indirect dependency.
As soon as we directly use something we should explicitly require the package.

.storybook/drupal-attribute.js
```diff
@@ -0,0 +1,56 @@
...
+          continue;
+        }
+
+        value = filtered.join(" ");
+      }
+
+      if (value === key) {
+        components.push(key);
+        continue;
+      }
```
donquixote: I think we don't need/want this block.
In Drupal we simply get `xyz="xyz"` for this case.
```suggestion
```

.storybook/drupal-attribute.js
```diff
@@ -0,0 +1,56 @@
...
+
+      if (value === "") {
+        components.push(`${key}=""`);
+        continue;
+      }
+
+      components.push(`${key}="${String(value)}"`);
+    }
+
+    return components.length ? ` ${components.join(" ")}` : "";
```
donquixote: The parent code does string concat aka `' ' + ..`, no need to change it here.. but ok.

tools/test-utils/index.js
```diff
@@ -1,22 +1,192 @@
...
+const { DrupalAttribute: BaseDrupalAttribute } = require("drupal-attribute");
 const twing = require("../../.storybook/environment");
```
AaronGilMartinez: Is it twing ot twig?

src/compositions/bcl-attributes/attributes.html.twig
```diff
@@ -0,0 +1,97 @@
...
+%}
+
+{% set attr_array = create_attribute()
+  .setAttribute('data-case', 'array')
+  .setAttribute('data-list', ['one', 'two', null, 'three', ''])
+%}
+
+{% set attr_array_empty_values = create_attribute()
+  .setAttribute('data-case', 'array-empty-values')
+  .setAttribute('data-list', ['one', '', null, ''])
```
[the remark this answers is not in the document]
tibi2303: added, thanks

PR #747, opened 2026-03-20
OEL-4605: Replaced ss-1 and ss-2 with ss-3.

tibi2303:


src/components/bcl-select/select.html.twig
```diff
@@ -123,7 +125,14 @@
 {%- if _options is not empty and _options is iterable %}
   {%- for _option in _options %}
     {%- if _option.optgroup is defined -%}
-      <optgroup label="{{ _option.label }}">
+      <optgroup
+        label="{{ _option.label }}"
+        {% if _option.select_all is defined and _option.select_all %}
```
donquixote: Can't we do this?

```suggestion
        {% if _option.select_all ?? false %}
```

src/components/bcl-select/select.html.twig
```diff
@@ -123,7 +125,14 @@
...
+          data-selectall="true"
+        {% elseif _select_by_group %}
+          data-selectall="true"
+        {% endif %}
```
donquixote, line 132: The two cases have the identical outcome. Is it intended?

donquixote: If so, we could use a single condition with "or".

src/themes/default/src/scss/_multiselect.scss
```diff
@@ -78,120 +204,372 @@ select.multi-select {
         }
       }
```
donquixote, line 206: Funny, the section above is the only part remaining from this file?
Intended?

src/themes/default/src/scss/_multiselect.scss
donquixote, line 1: A lot of code, hard to review.

donquixote: Given that we already replace most of this file, should we name it `slim-select.scss`?
Or is this about more than slim select?

src/themes/default/package.json
```diff
@@ -31,8 +31,7 @@
     "flag-icons": "7.5.0",
     "glob": "13.0.6",
     "npm-run-all": "4.1.5",
-    "slim-select": "1.27.1",
-    "slim-select-2": "github:tibi2303/slim-select#master"
+    "slim-select": "3.4.3"
```
donquixote, line 34: Is there a reason why most of the dependency versions here are fixed instead of using version ranges?
(I am not asking you to change it, it is just a question)

package.json
```diff
@@ -128,16 +127,14 @@
...
     "path-browserify": "^1.0.1",
     "postcss": "^8.5.6",
     "postcss-scss": "4.0.9",
     "prettier": "3.8.1",
     "process": "^0.11.10",
     "rimraf": "6.1.3",
     "size-limit": "12.0.0",
-    "slim-select": "1.27.1",
-    "slim-select-2": "github:tibi2303/slim-select#master",
+    "slim-select": "3.4.3",
```
donquixote, line 137: Also here in root package.json I see fixed versions instead of `^3.4.3`, also for other packages.

src/themes/default/src/scss/_multiselect.scss
donquixote: I see now the changes are a lot less if we compare to v2 :)

donquixote: I changed the commit history a bit so that git now understands the history as coming from v2.

donquixote: compare these two git blame views:
- https://github.com/openeuropa/bootstrap-component-library/blame/5c4452fc059db9056dd5fcec26e252fb4e62ffcf/src/themes/default/src/scss/_multiselect.scss (with split history)
- https://github.com/openeuropa/bootstrap-component-library/blame/13d83f821444bb6f7a7a8ee75309da559a4923cb/src/themes/default/src/scss/_multiselect.scss (single commit)

donquixote: Otherwise I trust you with that scss.
If UAT is good in the end, I won't complain :)

donquixote (approved):
I am approving. But please review my push:
- I rebased on latest `origin/master`
- I split the commit for continuity of history of v2 scss file.

src/themes/default/src/scss/_multiselect.scss
tibi2303: ok, great, thanks, i needed to fix a conflict though, didnt rebased only merged dev into this branch

src/themes/default/package.json
tibi2303: I usually update using upgrade-interactive command when we had yarn, i guess now pnpm update --interactive because i have control over the versions used.. i usually do it monthly

package.json
tibi2303: see the other comment, thanks

src/components/bcl-select/select.html.twig
tibi2303: thanks, indeed

src/components/bcl-select/select.html.twig
tibi2303: done, simplified both conditions into one

PR #750, opened 2026-04-09
OEL-4672: Fix most feasible audit dev dependency.

tibi2303:


donquixote:
It seems alright, but what specifically should I be looking for? Some before vs after check?

donquixote (approved):

donquixote:
It looks like an improvement, we can improve more later.

PR #751, opened 2026-04-10
OEL-4643: Media Copyright Accessibility improvements

piotrsmykaj:


src/components/bcl-card/card.html.twig
```diff
@@ -157,16 +161,22 @@
...
+        src="{{ _image.path }}"
+        alt="{{ _image.alt }}"
+        class="{{ _img_class }}"
+        {%- if _image_description_id is not empty %} aria-describedby="{{ _image_description_id }}"{% endif %}
+      >
+      {%- if _image_description_id is not empty %}<span id="{{ _image_description_id }}" class="visually-hidden">{{- _image_copyright -}}</span>{%- endif %}
+      {%- if _image_copyright is not empty %}
+        <figcaption>
+          <div class="bcl-card__image-footer">
+            <div class="bcl-copyright">{%- if _image_copyright_label is not empty %}<span class="visually-hidden">{{- _image_copyright_label -}}</span>{% endif %}{{- _image_copyright -}}</div>
```
donquixote: Now we have the copyright label and the copyright text without a space between them?

src/components/bcl-card/card.html.twig
```diff
@@ -261,16 +271,22 @@
...
+          </div>
+        </figcaption>
+      {%- endif %}
+    </figure>
```
donquixote: I see code duplication with the horizontal and vertical version of card.
It was already the case before but now it is worse because the code has grown.
It could be mitigated by capturing that html snippet.
But it's a tradeoff if we want to do that.

src/components/bcl-card/card.html.twig
donquixote: I see the copyright printed twice. One time before the figcaption (invisible), one time inside of it (visible). Also in the first place I don't see the condition on whether copyright exists at all.

src/components/bcl-carousel/carousel.html.twig
```diff
@@ -119,7 +122,15 @@
     {% endif %}
     >
     {%- if _item.image is defined and _item.image is not empty %}
-      {{- _item.image }}
+      <figure class="m-0">
```
donquixote: This will wrap in `<figure>` even if there is no description. But that's ok I think. It is easier for styling and for the template.

src/components/bcl-carousel/carousel.html.twig
```diff
@@ -119,7 +122,15 @@
...
+        {{- _item.image }}
+        {%- if _item.description_id is defined and _item.description_id is not empty -%}
+          <span id="{{ _item.description_id }}" class="visually-hidden">{{- _item.copyright -}}{%- if _item.caption_title is defined and _item.caption_title is not empty %} {{- _item.caption_title -}}{% endif %}{%- if _item.caption is defined and _item.caption is not empty %} {{- _item.caption -}}{% endif %}</span>
+        {%- endif -%}
+        {%- if _item.copyright is defined and _item.copyright is not empty -%}
+          <figcaption class="bcl-copyright">{%- if _copyright_label is not empty %}<span class="visually-hidden">{{- _copyright_label -}}</span>{% endif %}{{- _item.copyright -}}</figcaption>
+        {%- endif -%}
+      </figure>
```
donquixote, line 158: We used to have copyright below the caption, and now we have it below the image?

src/components/bcl-card/card.html.twig
piotrsmykaj: good point, fixed

src/components/bcl-card/card.html.twig
piotrsmykaj: Created a macro to wrap the duplicated code

src/components/bcl-card/card.html.twig
piotrsmykaj: fixed

src/components/bcl-carousel/carousel.html.twig
piotrsmykaj: Now the image, caption wrapper and copyright figcaption were wrapped in the figure element

src/components/bcl-card/card.html.twig
```diff
@@ -59,6 +60,28 @@
 {% set _content = content|default('') %}
 {% set _image = image|default({}) %}
 {% set _image_copyright = image_copyright|default('') %}
+{% set _image_copyright_label = image_copyright_label|default('') %}
+{% macro image_figure(image, img_class, copyright, copyright_label) %}
+  <figure class="m-0">
```
donquixote: We could add a class on the `<figure>` tag to allow to target it.
Maybe `bcl-card__image-wrapper`?
-> to be discussed with @tibi2303

src/components/bcl-card/card.html.twig
```diff
@@ -59,6 +60,28 @@
...
+    <img
+      src="{{ image.path }}"
+      alt="{{ image.alt }}"
+      class="{{ img_class }}"
+    >
+    {%- if copyright is not empty %}
+      <figcaption>
+        <div class="bcl-card__image-footer">
```
donquixote: We could combine these tags like

```suggestion
      <figcaption class="bcl-card__image-footer">
```
The question is if in other places we do need the two separate tags, in that case there could be an argument to keep them separate for consistency.

Semantically it seems that the figcaption _is_ the "image footer".

donquixote:
Could we add some copyright labels in the .story.js files so it appears in netlify and in local preview?

src/compositions/bcl-banner/banner.html.twig
```diff
@@ -93,6 +95,6 @@
   </div>{# -#}
 </div>
   {%- if _copyright is not empty -%}
-    <div class="bcl-copyright">{{- _copyright -}}</div>
+    <div class="bcl-copyright">{%- if _copyright_label is not empty %}<span class="visually-hidden">{{- _copyright_label -}}</span>{% endif %}{{- _copyright -}}</div>
```
donquixote: I wonder if we need a space between the visually-hidden span and the actual copyright, given that `<span>` is an inline tag.
This also applies for other templates with this element.

src/components/bcl-carousel/carousel.html.twig
```diff
@@ -119,36 +120,43 @@
...
+      <figure class="m-0">
+        {{- _item.image }}
+        {%- if (_item.caption is defined and _item.caption is not empty) or (_item.link is defined and _item.link is not empty) or (_item.caption_title is defined and _item.caption_title is not empty) %}
+          {%- set _caption_classes = 'carousel-caption' %}
+          {%- if _item.caption is not empty and _item.link is empty %}
+            {%- set _caption_classes = _caption_classes ~ ' d-none d-md-block' %}
+          {%- endif -%}
+          <div
+            class="{{ _caption_classes }}"
+          >
```
donquixote, line 132: Are we printing this caption wrapper even if there is no caption?

src/components/bcl-carousel/carousel.html.twig
```diff
@@ -119,36 +120,43 @@
...
+            {%- if _item.caption is not empty or _item.caption_title is not empty %}
+              {%- if _item.caption_classes is defined and _item.caption_classes is not empty -%}
+                <div class="{{ _item.caption_classes }}">
+              {%- endif %}
```
donquixote, line 136: I'm a bit confused because we have `_caption_classes` and then we have `_item.caption_classes`.
But I see that this was already the case before, so up to @tibi2303 to have a second look.

src/components/bcl-card/card.html.twig
tibi2303: not sure of the use cases when we will target this for now but i think it's a good idea.

src/components/bcl-card/card.html.twig
tibi2303: if the styles are the same, yeah, it could be a small cleanup

src/compositions/bcl-banner/banner.html.twig
tibi2303: i think it still makes a difference between element spacing, if it's necessary we could keep it like this

src/components/bcl-carousel/carousel.html.twig
tibi2303: actually i think most of the use cases are with caption - until now we didnt had any announce of this, we can still have a foresight and add the if

src/components/bcl-carousel/carousel.html.twig
tibi2303: i think here is indeed confusing the naming, but this 2 _caption_classes make 2 different things. If we rename it for item, we need bc. So let's keep it like this, maybe create a follow up ticket for it.

src/components/bcl-card/card.html.twig
tibi2303: I agree with this kind of changes but this tends to overextend the ticket.

src/compositions/bcl-banner/banner.html.twig
donquixote: In other places, Piotr already added a space so that it would be e.g. `<span class="visually-hidden">Image credit: </span>Pablo Picasso`.

src/components/bcl-carousel/carousel.html.twig
donquixote: my point is that we have a condition below about whether caption is empty, but this print is outside of that.

src/compositions/bcl-banner/banner.html.twig
piotrsmykaj: let's add it.

src/components/bcl-carousel/carousel.html.twig
piotrsmykaj: @donquixote the "endif" above the `<div>` closes only the inner "if", that adds the "d-none d-md-block" classes. The outer "if" (line 125), wraps the entire block, including the `<div>`, and it closes at line 149. I don't see a bug here.

donquixote: Ah, so the `<div>` also wraps `_item.link` or `_item.caption_title`, not just `_item.caption`.

src/components/bcl-carousel/carousel.html.twig
donquixote: The caption should also be wrapped in the `<figcaption>` tag for semantic reasons.
(confirm with Dora if you want)

src/compositions/bcl-banner/banner.html.twig
```diff
@@ -93,6 +95,11 @@
   </div>{# -#}
 </div>
   {%- if _copyright is not empty -%}
-    <div class="bcl-copyright">{{- _copyright -}}</div>
+    <div class="bcl-copyright">
+      {%- if _copyright_label is not empty -%}
+        <span class="visually-hidden">{{- _copyright_label }} </span>
+      {%- endif -%}
+      {{- _copyright -}}
+    </div>
```
donquixote, line 103: Given this is for a background image, not an image tag, we cannot use `<figure>` here. But we still add the `visually-hidden` label.
-> Good.

Maybe there could be ways to improve this, but let's keep it like this until we get a new request.

src/components/bcl-carousel/carousel.html.twig
donquixote: UPDATE: Check with others before you change this, I don't want to cause unnecessary back and forth.

donquixote: Ok let's go with this version until we get different request.

donquixote (approved):
Will need a rebase before merging.

donquixote (approved):

PR #760, opened 2026-05-20
OEL-4762: Updates on packages.

tibi2303:


.github/workflows/ci.yml
```diff
@@ -70,7 +67,6 @@ jobs:
       - uses: actions/checkout@v6
       - uses: pnpm/action-setup@v4
         with:
-          version: 10.6.2
```
donquixote, line 73: If we don't fix the version it will always use latest?

.github/workflows/dependency_review.yml
donquixote, line 1: What does it do?

patches/locutus@3.0.36.patch
donquixote, line 1: I don't know what it does, but I trust you :)

donquixote (approved):
Approving, with questions for curiosity.

.github/workflows/ci.yml
tibi2303: yes, which i want here. I want pipeline failing if is not working with latest.

.github/workflows/dependency_review.yml
tibi2303: It runs actions/dependency-review-action@v5 on every PR and fails if a dependency change introduces any vulnerability at low or higher, including both runtime and development scopes.

donquixote (approved):

PR #765, opened 2026-05-28
OEL-4774: Dynamic output for templates folder structure.

tibi2303:


src/themes/default/bcl-builder.config.js
```diff
@@ -39,8 +39,17 @@ const toOutputRelativePath = (targetPath) =>
   path.relative(outputFolder, targetPath);
 const toOutputRelativePackagePath = (packagePath, ...segments) =>
   toOutputRelativePath(path.resolve(packagePath, ...segments));
+const getCopyUpToGlobBase = (sourceGlob) =>
+  path
+    .normalize(sourceGlob.split("**")[0])
+    .replace(/[\\/]+$/, "")
+    .split(/[\\/]+/).length;
```
donquixote, line 46: So we are counting path fragments in a "normalized" wildcard path before the `'**'` part.
I see this being called once.

src/themes/default/bcl-builder.config.js
```diff
@@ -39,8 +39,17 @@ const toOutputRelativePath = (targetPath) =>
...
 const bootstrapIconsPath = resolvePackagePath("bootstrap-icons");
 const twigTemplatesPath = resolvePackagePath("@openeuropa/bcl-twig-templates");
+const twigTemplatesGlob = toOutputRelativePackagePath(
+  twigTemplatesPath,
+  "templates/**/*.twig",
+);
```
donquixote, line 52: So here we are just moving an expression up that was inline before, because we want to use the variable again.

Then we calculate number of levels up as `getCopyUpToGlobBase(twigTemplatesGlob)`.

From what I see this would depend only on `twigTemplatesPath`, so whatever we get from `resolvePackagePath("@openeuropa/bcl-twig-templates");`.

What I don't get is why would this number change between versions?
Maybe because `@openeuropa/bcl-twig-templates` is found in different places?

donquixote: What would be a typical value of `twigTemplatesPath`?

src/themes/default/bcl-builder.config.js
```diff
@@ -219,11 +228,9 @@ module.exports = {
       options: { up: 7 },
     },
     {
-      from: [
-        toOutputRelativePackagePath(twigTemplatesPath, "templates/**/*.twig"),
-      ],
+      from: [twigTemplatesGlob],
       to: path.resolve(outputFolder, "templates"),
-      options: { up: 5 },
+      options: { up: getCopyUpToGlobBase(twigTemplatesGlob) },
```
donquixote, line 233: So the copy command gets a `from:` path (wildcard), a `to:` path, and a `options.up:` number.
What does this `options.up:` number do?
Can you explain this with example paths?
E.g.
- `twigTemplatesGlob` is `path/to/templates/**/*.twig`.
- one matching template is `path/to/templates/component/accordion/accordion.twig`.
- `path.resolve(outputFolder, "templates")` is `/something/dest`?
- `options.up` is 5 (as example)
- -> where does `accordion.twig` get copied?

donquixote (commented):
Trying to understand better..

src/themes/default/bcl-builder.config.js
tibi2303: Yes. copyfiles needs a numeric up value: it removes that many path fragments from each matched source file before writing to the destination.
The source glob is built from the actual resolved @openeuropa/bcl-twig-templates path, so the number of fragments before ** can differ depending on install layout. In this checkout it is:
../../../node_modules/@openeuropa/bcl-twig-templates/templates/**/*.twig
Counting those fragments gives 7, which strips the full package/template prefix and leaves:
bcl-accordion/accordion.html.twig
It is only called once now, but the helper is there to make the intent explicit and avoid another hard-coded up: 5 / up: 7 drift if the resolved package path changes again.

src/themes/default/bcl-builder.config.js
tibi2303: Yes, exactly. twigTemplatesGlob was pulled into a variable because we need the same resolved glob in two places: from and the up calculation.

A typical value of twigTemplatesPath depends on the package manager/install layout. In this repo checkout I see:

/Users/.../bootstrap-component-library/node_modules/@openeuropa/bcl-twig-templates

So twigTemplatesGlob, relative to src/themes/default, becomes:

../../../node_modules/@openeuropa/bcl-twig-templates/templates/**/*.twig
That needs up: 7.

The number can change if @openeuropa/bcl-twig-templates resolves from a different physical location, for example a workspace/symlink target, pnpm store path, npm package layout, or another monorepo install shape. The old hard-coded value only worked for one specific path shape. This change ties up to the actual resolved glob used by copyfiles, so both stay in sync.

src/themes/default/bcl-builder.config.js
tibi2303: options.up is a copyfiles option. It removes that many leading path fragments from the matched source file path before appending the remainder to the destination directory.
```
from glob:
path/to/templates/**/*.twig

matched file:
path/to/templates/component/accordion/accordion.twig

to:
 /something/dest

up: 5

Remove the first 5 fragments:

accordion.twig
So the result is:

/something/dest/accordion.twig
```

src/themes/default/bcl-builder.config.js
donquixote: Ok so the difference is not between versions of bcl, but depending where and how the operation runs.

donquixote (approved):
I think it's a bit confusing with the relative paths + counting of fragments.
We might improve this in the future to be more direct and have less confusing function names and var names.

PR #767, opened 2026-06-03
OEL-4810: Fix generated asset copying with pnpm layout.

### LINKED ISSUE #762: @openeuropa/bcl-bootstrap 1.12.0+: individual icon SVGs missing from published npm package
Starting from v1.12.0, the @openeuropa/bcl-bootstrap npm package no longer includes individual Bootstrap icon SVG files in the icons/ directory. Only the custom icons/twitter.svg replacement is published. 

This breaks the bcl-builder sprite task for any consumer project.

Please compare https://unpkg.com/browse/@openeuropa/bcl-bootstrap@1.11.0/icons/ with  https://unpkg.com/browse/@openeuropa/bcl-bootstrap@1.12.0/icons/

tibi2303:


tibi2303:
Ok, I improved most of the generation script and included tests for each of them. Also a summary testing regarding the count of files was added.

package.json
```diff
@@ -20,7 +20,7 @@
     "changelog": "lerna-changelog",
     "clean": "rimraf build assets",
     "clean:themes": "lerna --scope '@openeuropa/bcl-theme-*' run clean",
-    "copy:assets": "copyfiles -f src/themes/**/*min.* node_modules/slim-select/dist/slimselect.js static/* assets/ && mv -f assets/slimselect.js assets/slimselect.min.js && copyfiles -u 3 src/themes/**/fonts/* assets && copyfiles -u 3 src/themes/**/fonts/files/* assets",
+    "copy:assets": "node scripts/copy-assets.js",
     "copy:index": "copyfiles -f static/*.html static/*.ico build",
```
AaronGilMartinez, line 24: This is curiosity: why custom code in favor of a package? I see that we still use the package but we have replaced it in some parts.

src/themes/joinup/package.json
```diff
@@ -7,8 +7,7 @@
   "scripts": {
     "align-templates": "lerna --scope \"@openeuropa/bcl-twig-templates\" run prepublish",
     "build": "npm-run-all 'build:*'",
-    "build:fonts-css": "copyfiles -f ../../../node_modules/@fontsource/roboto/{lat*,cyr*,gree*}-{400,500,700}*.css fonts",
-    "build:font-files": "copyfiles -u 6 ../../../node_modules/@fontsource/roboto/files/roboto-{latin*,cyrillic,greek}-{400,500,700}* fonts",
```
AaronGilMartinez, line 11: I see that the command was copying multiple files but now I see only one font and size in the test, same for icons. 

Now I'm thinking a way of creating snapshots of current files like having multiple commands:

- Test: verifies that the list of generated files in assets is correct.
- Update: updates list of files to be checked in the tests given current list of files.

But I wouldn't do this in the ticket, I would create a separate task after a bit of analysis.

AaronGilMartinez (changes requested):

package.json
tibi2303: Copyfiles is still used to copy small simple files, not what we've tried to do here. And i did this because the command was hard to read and got very complex. Also pnpm depends on where the files are relocated and the paths are different from npm.

src/themes/joinup/package.json
tibi2303: I gave it a try and added tests for all the files we care about - the ones that are released.

AaronGilMartinez (approved):

PR #769, opened 2026-06-10
OEL-4822: Upgrade eslint from 8 to 10 and adapted.

tibi2303:


package.json
```diff
@@ -106,17 +105,10 @@
     "browser-update": "3.3.63",
     "buffer": "^6.0.3",
     "copyfiles": "2.4.1",
-    "eslint": "^8.57.1",
-    "eslint-config-airbnb": "19.0.4",
-    "eslint-config-airbnb-base": "15.0.0",
+    "eslint": "^10.4.1",
     "eslint-config-prettier": "10.1.8",
```
AaronGilMartinez: Eslint-config-prettier and eslint-plugin-prettier remain in package.json but are never referenced in eslint.config.cjs. Either integrate them (e.g., add prettier flat config) or remove them.

bootstrap/package.json
```diff
@@ -24,7 +24,7 @@
     "bootstrap": "5.3.8",
     "bootstrap-icons": "1.13.1",
     "copyfiles": "2.4.1",
-    "eslint": "^8.57.0",
+    "eslint": "^10.4.1",
```
AaronGilMartinez: Is eslint package needed at this level?

scripts/watch.js
```diff
@@ -5,7 +5,7 @@ const browsersync = require("browser-sync");
 const path = require("path");
 const instance = process.argv.slice(3)[0] || "default";
 const ports = {
-  dev: "5000",
+  dev: "5003",
```
AaronGilMartinez: Is this change intended? if so the readme needs update.

AaronGilMartinez (changes requested):

scripts/watch.js
tibi2303: i have problems on my mac and i usually change that to have live preview.. forgot to revert it

package.json
tibi2303: removed them, indeed, thanks

bootstrap/package.json
tibi2303: indeed, is already at root level. i removed it

AaronGilMartinez (approved):

AaronGilMartinez (approved):

PR #770, opened 2026-06-10
OEL-4820: Fix visible corners on card image.

### LINKED ISSUE #439: Card v2 - overflow not hidden
### Description
https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/resources/sass/components/_card_v2.scss#L38
Can you please set `overflow:hidden;` to the main `.card.v2` ? Not only to the `.horizontal` variant ?

### How to reproduce it
Put an image without header. The top corners are not round

tibi2303:


src/themes/default/src/scss/_card.scss
```diff
@@ -1,5 +1,6 @@
 .card {
   border-color: $gray-300;
+  overflow: hidden;
```
AaronGilMartinez: This may be too broad and break other than v2.

AaronGilMartinez (changes requested):

src/themes/default/src/scss/_card.scss
tibi2303: indeed, it was. I changed the fix for a more backward compatible approach.

src/themes/default/src/scss/_card.scss
```diff
@@ -1,5 +1,10 @@
 .card {
   border-color: $gray-300;
+  .bcl-card__image-wrapper:first-child {
+    overflow: hidden;
+    border-top-left-radius: var(--bs-card-border-radius);
+    border-top-right-radius: var(--bs-card-border-radius);
+  }
```
AaronGilMartinez: https://preview-770--oelibrary.netlify.app/default/?path=/story/components-card--horizontal

Seems to apply to top right here

tibi2303: i directly selected that because on horizontal is inside a div.

AaronGilMartinez (approved):

piotrsmykaj (approved):

PR #780, opened 2026-08-14
OEL-4973: Harden BCL dependency installation and release workflow.

tibi2303:


README.md
AaronGilMartinez, line 1: The documentation still lags behind the migration because builder/README.md:8 tells users to run npm run ecl-builder scriptName, and README.md:27 still uses npm install as the main setup example instead of a pnpm-first flow.

AaronGilMartinez (changes requested):

README.md
tibi2303: updated docs and readmes.

AaronGilMartinez (approved):

PR #784, opened 2026-08-25
OEL-4984: Slim select 2 cleanup and backward compatibility.

tibi2303:


pnpm-workspace.yaml
```diff
@@ -112,6 +112,8 @@ allowBuilds:
 
 patchedDependencies:
   locutus@3.0.36: patches/locutus@3.0.36.patch
+  # Keep ARIA state on the focused search combobox and complete keyboard support.
```
piotrsmykaj, line 115: I would suggest describing in more detail (perhaps as a list?) what the patch does.

PR #785, opened 2026-08-25
OEL-4961: Increase icon size for Facts and figures

piotrsmykaj:


src/compositions/bcl-fact-figures/fact-figures.html.twig
```diff
@@ -90,7 +90,7 @@
         {%- set _item = _item|merge({
           icon: _item.icon|merge({
             path: _icon_path,
-            size: 'l',
+            size: '3xl',
```
tibi2303: i think this should not change.. the impact is big..

I think we need to implement a new parameter, like feature_mode. where we set this icon size and the text alignment - same as in figma.

tibi2303: rest is fine

piotrsmykaj: as agreed I changed the approach

tibi2303 (approved):

src/compositions/bcl-fact-figures/fact-figures.html.twig
```diff
@@ -36,6 +38,7 @@
 {% set _equal_height = equal_height ?? false %}
 {% set _cols_extra_classes = cols_extra_classes|default('') %}
 {% set _icon_path = icon_path|default('') %}
+{% set _icon_size = icon_size|default('3xl') %}
```
AaronGilMartinez, line 41: Why the change of default?

AaronGilMartinez (changes requested):

src/compositions/bcl-fact-figures/fact-figures.html.twig
piotrsmykaj: 3xl is the new default icon size for the fact & figures, and L is kept only for backward compatibility, which we handle at the oe_bootstrap_theme level.

AaronGilMartinez: True I mixed icon component with fact and figures.

AaronGilMartinez (approved):

PR #789, opened 2026-09-14
OEL-5010: Fix Slim Select option announcements and remove button accessibility.

tibi2303:


patches/slim-select@3.6.1.patch
AaronGilMartinez, line 1: How was this patch produced? Is there an slim-select PR or issue for these accessibility fixes? If yes, please link it in the patch header so we have a clear trigger to drop the patch on a future upgrade. If no, should we open one?

AaronGilMartinez (changes requested):

patches/slim-select@3.6.1.patch
tibi2303: they already moved to slim select 4 on master and i think they will not accept the pr. I can still open it to make it easier to read the patch.

AaronGilMartinez: We can apply the patch but this is an issue, we are working on a non supported version.

AaronGilMartinez (approved):

PR #790, opened 2026-09-17
OEL-5028: Add optional tone scale to color scheme.

julien-:


src/themes/default/src/scss/color_scheme/_tones.scss
```diff
@@ -0,0 +1,35 @@
...
+
+// Returns the color with a new OKLCH lightness, mapped into sRGB.
+@function tone-color($color, $lightness) {
+  $tone: color.to-gamut(
+    color.change($color, $lightness: $lightness, $space: oklch),
+    $space: rgb,
+    $method: local-minde
+  );
+
+  @return rgb(
```
tibi2303, line 20: Could we output hex here regardless of minification? With minify: false, this produces rgb(...), while the acceptance criteria require hex.

builder/README.md
```diff
@@ -86,6 +86,50 @@ We've also added some new variables that overwrite and make use of extra classes
 - `.text-color-default`: Text color is changed based on the 'text' variable, overriding all text colors inside a component.
 - `.bg-default`: Background color is changed based on the 'background' variable. It adds a background for the component but does not override if it already has a `bg` class (e.g., `bg-primary`).
 
+#### Color tones
+
+Optional. Set `$color-scheme-tones` in the entry file to get a tone scale (lighter / darker variants) for each scheme color.
+The map key is the tone name, the value is the [OKLCH](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch) lightness.
+The hue is kept and the chroma is reduced only when the color is out of sRGB.
+OKLCH gives the same visible lightness for all colors, so the text contrast ([WCAG](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html)) stays about the same across colors.
```
tibi2303, line 94: Could we soften this claim? Equal OKLCH lightness can still give different WCAG pass/fail results. I’d suggest: “OKLCH provides approximately uniform perceptual lightness. Each foreground/background pair still needs a contrast check.”

src/themes/default/src/scss/color-scheme.scss
```diff
@@ -13,3 +13,4 @@
 @import "color_scheme/list_group";
 @import "color_scheme/utilities";
 @import "color_scheme/badges";
+@import "color_scheme/tones";
```
tibi2303, line 16: Looks good overall. I checked that the default CSS stays unchanged and exclusions work. A small regression test covering the default output and tone generation would be useful.

src/themes/default/src/scss/color_scheme/_tones.scss
```diff
@@ -0,0 +1,35 @@
...
+// Color scheme - Tones
+
+// Tone name => OKLCH lightness. Empty map: no tone is generated.
+$color-scheme-tones: () !default;
+
+// Scheme keys without tones.
+$color-scheme-tones-exclude: "text", "background" !default;
+
+// Returns the color with a new OKLCH lightness, mapped into sRGB.
+@function tone-color($color, $lightness) {
```
tibi2303, line 13: Could we make chroma, hue, and the gamut-mapping method optional parameters for the tone generator? That would let projects reproduce their Figma palettes while keeping the current lightness-only behavior as the default. We’ve confirmed that 13% / 0.17 / 268° with clip matches the Figma swatch #0C004B we have for oel.

